PR #3, opened 2020-10-16
Add class::max_autoconn configuration

edk0:


glguy (approved):
This is a straightforward code change, but it's a bit unsatisfying that we have servers and clients sharing the `class` configuration section without sharing all the settings.

I'll grant that things aren't necessarily getting worse as previously we had a configuration parameter shared across two different interpretations.

Would be worth either checking that either `max_autoconn` or `max_number` were not specified if they were not going to be used? This would be especially good for someone with an old configuration file now surprised that the `max_number=1;` setting was now being ignored in his server configuration. Perhaps there are already a few different settings that are ignored in one case or the other for clients/servers and so this isn't the time to solve that! In any case that's my feedback on the topic :)

edk0:
`max_number` will still be used as a maximum, just not specifically for autoconnect, so (a) we could probably check that `max_number >= max_autoconn` and (b) if you didn't get that it's probably worth documenting.

In general there are lots of class attributes that are shared but only meaningful for one thing, which is worth fixing at some point.

glguy (approved):

PR #4, opened 2020-10-16
Implement the solanum.chat/identify-msg vendor cap

edk0:


modules/core/m_identified.c
```diff
@@ -0,0 +1,64 @@
...
+
+static void m_identified(struct MsgBuf *msgbuf, struct Client *client_p, struct Client *source_p, int parc, const char *parv[])
+{
+	struct Client *target_p = find_person(parv[1]);
+	const char *nick = parv[2];
+
+	if (target_p == NULL)
+		return;
+
+	if (irccmp(target_p->name, nick))
```
glguy, line 38: If this UID is _not_ using the nickname that services says its identified for, should this set this client to being unidentified?

glguy (approved):
This seems like useful functionality and assuming my comments don't cause any concern the code looks good.

One request would be to document the behavior of this extension in English somewhere. If it is already we could provide a URL to that documentation in the implementation.

include/client.h
```diff
@@ -494,6 +495,9 @@ struct ListClient
 #define IsTGExcessive(x)	((x)->flags & FLAGS_TGEXCESSIVE)
 #define SetTGExcessive(x)	((x)->flags |= FLAGS_TGEXCESSIVE)
 #define ClearTGExcessive(x)	((x)->flags &= ~FLAGS_TGEXCESSIVE)
+#define IsIdentified(x)		((x)->flags & FLAGS_IDENTIFIED)
```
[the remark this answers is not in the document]
edk0: It's alright in the file, just wrong in the diff view, somehow.

modules/core/m_identified.c
edk0: I don't think it matters much. If they're not using the right nickname because they changed nick, they've already been unidentified, and that's the only case I can think of.

edk0:
> One request would be to document the behavior of this extension in English somewhere. If it is already we could provide a URL to that documentation in the implementation.

Yeah, I want to make `solanum.chat/{identify-msg,identified}` a URL to the documentation, but there isn't any yet and we don't have the requisite website machinery either. I'll make an issue on the website repo to track it.

PR #6, opened 2020-10-16
extb_canjoin: disable inside +e

edk0:


dwfreed:
This should also include +I (`CHFL_INVEX`)

edk0:
Unlike +e, I don't think there are any bad things unlocked by nesting bans inside invex lists

jesopo (approved):

dwfreed (approved):

PR #11, opened 2020-10-18
m_cap: send multiline cap lists to everyone

edk0:
IRCv3 doesn't spec this for non-302 CAP LS, but as far as I can tell this is just a mistake. The original cap implementation always did it.

### DIFF
--- a/modules/m_cap.c
+++ b/modules/m_cap.c
@@ -198,11 +198,6 @@ clicap_generate(struct Client *source_p, const char *subcmd, int flags)
 					data != NULL ? data : "") < 0
 					&& buf_list[0] != '\0') {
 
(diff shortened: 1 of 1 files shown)


dwfreed:
From IRC:

```
<dwfreed> honestly? giving the 3.1 caps priority so they show up first, and then not doing multiline for 3.1 clients at all seems like the best bet
<ghoulguy> Yeah, we just need old clients to work, not work great
```

3.1 clients would need to blindly request newer capabilities that didn't make the list, which sucks, but is the best we've got.  Chances are those are going to be handled by scripts/plugins rather than the client implementation itself anyway, so that is presumably easier to accomplish.

PR #12, opened 2020-10-18
m_names: honour userhost-in-names for global NAMES

edk0:


modules/m_names.c
```diff
@@ -172,14 +172,33 @@ names_global(struct Client *source_p)
...
 			continue;
 
-		if((cur_len + NICKLEN + 2) > (BUFSIZE - 3))
+		if (IsCapable(source_p, CLICAP_USERHOST_IN_NAMES))
 		{
-			sendto_one(source_p, "%s", buf);
-			cur_len = mlen;
-			t = buf + mlen;
+			/* !@, space */
+			if (cur_len + strlen(target_p->name) + strlen(target_p->username) + strlen(target_p->host) + 3 >= BUFSIZE - 3)
```
dwfreed, line 178: As mentioned on IRC, compilers will convert strlen of string literals into constants, so we can avoid using the magic constants ourselves:

```suggestion
			if (cur_len + strlen(target_p->name) + strlen(target_p->username) + strlen(target_p->host) + strlen(" !@") >= BUFSIZE - 3)
```

modules/m_names.c
```diff
@@ -172,14 +172,33 @@ names_global(struct Client *source_p)
...
+				cur_len = mlen;
+				t = buf + mlen;
+			}
+
+			tlen = sprintf(t, "%s!%s@%s ", target_p->name, target_p->username, target_p->host);
+		}
+		else
+		{
+			/* space */
+			if(cur_len + strlen(target_p->name) + 1 >= BUFSIZE - 3)
```
dwfreed, line 191: As above:

```suggestion
			if(cur_len + strlen(target_p->name) + strlen(" ") >= BUFSIZE - 3)
```

dwfreed (changes requested):

dwfreed (approved):
Good enough, until the One True Multiline comes along

PR #13, opened 2020-10-18
Add extensions/cap_realhost

edk0:
Implements the `solanum.chat/realhost` client capability, which sends the `solanum.chat/realhost` and `solanum.chat/ip` vendor tags on user-sourced messages as appropriate.

Hidden IPs and hostnames are sent to `auspex:hostname` opers. For consistency, if a remote oper wouldn't see an IP (because it's behind an IP spoof and not propagated), it's not shown to local opers either. Internally the oper version of the tag is implemented with an invisible cap named `?oper_realhost`.

---

*originally:* I'm not terribly happy with the state of this, but improving it depends on having tags that are gated by privs, which may itself depend on changing how privsets work.

Ideally the cap would be available to all users, but just wouldn't reveal hidden information to the unprivileged.


dwfreed:
The ideal would be possible if the outbound_msgbuf hook included the target information.  Otherwise, seems fine to me.

edk0:
As it stands the hook is called just once for something like a channel message, even though it goes out to lots of users, and the send machinery takes care of formatting different concrete lines for people with distinct capability sets. I'd like to keep that, so we probably need to teach the sending stuff about privs.

ncoevoet:
Sigyn approves

jesopo (dismissed):

jesopo (approved):

extensions/cap_realhost.c
```diff
@@ -0,0 +1,131 @@
...
+{
+	hook_data *data = data_;
+	struct MsgBuf *msgbuf = data->arg1;
+
+	if (data->client == NULL || !IsPerson(data->client))
+		return;
+
+	if (!IsIPSpoof(data->client) && !EmptyString(data->client->sockhost) && strcmp(data->client->sockhost, "0"))
+	{
+		if (IsDynSpoof(data->client))
```
glguy, line 72: Would this have been nicer as:

```c
		msgbuf_append_tag(msgbuf, "solanum.chat/ip", data->client->sockhost,
			IsDynSpoof(data->client) ? CLICAP_OPER_REALHOST : CLICAP_REALHOST);
```

Just to capture the fact that that's the only thing that depends on `IsDynSpoof`?

extensions/cap_realhost.c
```diff
@@ -0,0 +1,131 @@
...
+
+static void
+cap_realhost_umode_changed(void *data_)
+{
+	hook_data_umode_changed *data = data_;
+
+	if (!MyClient(data->client))
+		return;
+
+	if (data->client->localClient->caps & CLICAP_REALHOST)
```
glguy, line 90: Could we factor this update function out once and just use it in 3 places?

```c
		if (client->localClient->caps & CLICAP_REALHOST)
		{
			client->localClient->caps &= ~CLICAP_OPER_REALHOST;
			if (HasPrivilege(client, "auspex:hostname"))
				client->localClient->caps |= CLICAP_OPER_REALHOST;
		}
```

glguy: If I turn off `CLICAP_REALHOST` but I also had `CLICAP_OPER_REALHOST` then I'd keep `CLICAP_OPER_REALHOST`? Is it possible to turn caps off?

glguy (approved):

PR #14, opened 2020-10-18
m_dline: don't fail on /dline <time>

edk0:


glguy (approved):
This fix looks good, but I'm inclined to prefer doing the tests that `loc` is within `parc` at the point that `parv` is dereferenced.

```c
	dlhost = parv[loc];
	loc++;
```

It's at this point that we should be looking at `loc` to see if an IP address is available. Checking here matches the behavior below of checking for `"ON"` before using it and then advancing loc.

`tdline_time`'s potential use of `loc` is guaranteed to be safe because of the minimum parameter count up in the message handler entry, so it could remain unguarded.

edk0:
> It's at this point that we should be looking at `loc` to see if an IP address is available. Checking here matches the behavior below of checking for `"ON"` before using it and then advancing loc.

Like that? I think I prefer it this way too.

glguy (approved):
I like this version better, too. Thanks!

PR #15, opened 2020-10-18
extb_hostmask: handle banstr == NULL

edk0:


glguy (approved):
Yes, it's the case that an extban callback can be called with `banstr=NULL` when there's no ':' parameter , and matching on these masks asserts that they should not be `NULL`. Good fix.

PR #24, opened 2020-10-21
Don't track help indexes in git

edk0:
Remembering to check in regenerated ones every time we change the help is pointless extra work and creates pointless extra merge conflicts if there are two PRs with new help files in flight

No doubt this will conflict with #23, demonstrating the problem somewhat.


help/Makefile.am
```diff
@@ -1,7 +1,7 @@
 # Generated automatically from Makefile.in by configure.
```
glguy, line 1: Is this true? If so why are we modifying it? If not can we delete it?

help/Makefile.am
```diff
@@ -17,10 +17,12 @@ SYMLINKS=	topic accept cmode admin names links away whowas \
 
 all: index
 build:
-clean:
 depend:
 lint:
 
+clean:
```
glguy, line 23: We're adding a new `PHONY` target and there are a bunch of other PHONY targets. Normally I'd expect

```
.PHONY: all build clean depend lint index
```

Opinion?

glguy (changes requested):
If happy to be wrong on either of these changes.

help/Makefile.am
edk0: I think it's there so it shows up in the generated Makefile. The next line is definitely wrong, though.

help/Makefile.am
edk0: We're not adding it, just moving it down a few lines. I can't remember how but some of the autostuff generates a .PHONY in the Makefile proper.

glguy (dismissed):

glguy (approved):

PR #27, opened 2020-10-21
Document ITER_COMM_CHANNELS

edk0:


glguy (changes requested):
Could you document expectations about each parameter?

Some should not be modified by the macro user, some should be initialized, does any cleanup need to happen, etc.

glguy (approved):

PR #30, opened 2020-10-25
m_webirc: Blow away ident if we've got it

edk0:
Previously, anyone on the WEBIRC gateway would get its ident if it advertised one, which seems wrong to me. This allows ident to be used with WEBIRC while still letting users send their own ~username in band.

Also murder anything that sends WEBIRC after NICK/USER so we don't have confusing cases with undoing partial registrations. It's always been a protocol violation as well as a sanity violation to send WEBIRC after other stuff.


glguy (approved):
I agree that this feature implemented the behavior described.

I don't have an argument whether its a good or a bad idea yet, but if we want it I believe this implements it.

edk0:
I think I'm going to go ahead and land this today absent objections; I've pretty much convinced myself ident for webirc is both needlessly difficult to set up and largely incompatible with using ident to authenticate the webirc gateway itself (something m_webirc also tries to support). Once this is in, we can support gateway-validated ident using an option: `solanum.chat/ident=edk` or so.

PR #33, opened 2020-10-26
helpops: change umode char to h

edk0:
So it's incompatible with IP cloaking modules that use +h. We can choose between them. Or maybe just throw the IP cloaking modules out...


glguy (changes requested):
Change mention on line 223 and 232 not to be +H anymore?

glguy (approved):

PR #36, opened 2020-10-28
Update news

edk0:


glguy (approved):
While I haven't preformed a side-by-side comparison to the commit log for completeness, the changes here look good.

glguy (approved):
Better than ever! A scan through the top of git log suggests we have the big stuff now.

PR #38, opened 2020-10-28
Allow marking certain networks as secure via config

edk0:


glguy:
Should `UMODE_SSLCLIENT` become `UMODE_SECURE`?

glguy (approved):

PR #39, opened 2020-10-28
authproc: Don't try to notice dead clients

edk0:


glguy (approved):

glguy:
Alternate solution would have been to reinforce sendto_one_notice to perform this check so it can't be missed, but in discussion we established that there aren't many places clients can die.

PR #41, opened 2020-10-30
Fix a corner case of superset matching

edk0:
The algorithm we're using gets stuck if it has a ? and can only see a * to feed to it, even if it could skip over that * and consume a character following it. Remedy this by rearranging the input so * always precedes ? in runs of wildcards, so when we're matching ? we know we can skip things.

### DIFF
--- a/ircd/match.c
+++ b/ircd/match.c
@@ -104,6 +104,30 @@ int match(const char *mask, const char *name)
 	}
 }
 
+void
+match_arrange_stars(char *mask)
+{
+	char *swap = NULL;
+
+	for (char *p = mask; *p != '\0'; p++)
+	{
+		switch (*p)
+		{
+		case '*':
+			if (swap == NULL) break;
+			*swap++ = '*';
+			*p = '?';
+			break;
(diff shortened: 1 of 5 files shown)


ircd/match.c
```diff
@@ -104,6 +104,30 @@ int match(const char *mask, const char *name)
...
+match_arrange_stars(char *mask)
+{
+	char *swap = NULL;
+
+	for (char *p = mask; *p != '\0'; p++)
+	{
+		switch (*p)
+		{
+		case '*':
+			if (swap == NULL || swap == p) break;
```
glguy, line 117: Let's remove ` || swap == p`

glguy (approved):
I'm as confident about this logic as I'm going to get, and in particular I believe match_arrange_stars works correctly. I wouldn't mind a command explaining what property that function was imposing on its outputs. Based on the paper I've been looking over the transformation looks a lot like the linearization transform I've seen but I'm not confident in the connection.

glguy (approved):

PR #42, opened 2020-10-31
Remove shared blocks

edk0:


glguy:
Can we remove all the SHARED flag definitions in include/s_newconf.h ?

edk0:
Some of them are used for clusters too; implicit global clustering is also on my agenda, so I was thinking of leaving it for now and throwing them all out at once.

glguy (approved):
LGTM. I'd maybe inline `handle_remote_unresv` at this point, but even if we don't I'm good.

We talked on IRC that we can remove all the now-unused `#define` at the same time at a later date.

glguy (approved):

PR #43, opened 2020-10-31
Actually run tests with assertions enabled

edk0:
There's one that fails, which I'm remedying by adding a local client flag to shut it up. If it happens more often we might think about a generic mechanism for tests to silence/expect assertions.


glguy:
Do you expect that this flag will ever be used for anything else in the future? Could we put a comment on it explaining that it's actually for running tests and what kind of asserts we expect it to cancel out?

glguy (approved):

PR #44, opened 2020-11-01
Remove hardcoded TLSv1 disables

edk0:


glguy (approved):
Looks good. The mbedtls change feels non-obvious but I Googled around a bit and "3.1" is "TLSv1.0"... yay!

PR #47, opened 2020-11-05
Restructure channel mode handling, make it interact consistently with override

edk0:
The previous behaviour did one access check for a query, and then another one if you tried to set something, with the result that you couldn't see +eI lists with `MODE #foo e`, but you could by prefixing with a +: `MODE #foo +e`. Also you'd send an override snote for a query you could have done without override.

Most of this change is just making things more consistent; currently I've also got an override change in here to prevent using it to view +eI lists at all, but there's an argument to be had about how you should do that.

As a bonus, doing the parsing properly means we can restore the behaviour of `MODE #foo b x!y@z` (which once banned x!y@z, then didn't because override reauthorization hack).

I've been fairly conservative about rejecting invalid mode commands we can now detect before we start trying to effect them; see XXXs in the diff. There's a fair bit more we could do in that vein if we want to; I don't see any reason we couldn't process mode limits at parse time, for example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ ircd/version.c.last
 ssld/ssld
 wsockd/wsockd
 tests/core
+tests/chmode1
 tests/hostmask1
 tests/match1
 tests/msgbuf_parse1
--- a/extensions/chm_operonly_compat.c
+++ b/extensions/chm_operonly_compat.c
@@ -15,41 +15,33 @@ static const char chm_operonly_compat[] =
 static int _modinit(void);
 static void _moddeinit(void);
 static void chm_operonly(struct Client *source_p, struct Channel *chptr,
-	int alevel, int parc, int *parn,
-	const char **parv, int *errors, int dir, char c, long mode_type);
+	int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
 
 DECLARE_MODULE_AV2(chm_operonly_compat, _modinit, _moddeinit, NULL, NULL, NULL, NULL, NULL, chm_operonly_compat);
 
 static int
 _modinit(void)
 {
-	chmode_table['O'].set_func = chm_operonly;
-	chmode_table['O'].mode_type = 0;
-
+	chmode_table['O'] = (struct ChannelMode){chm_operonly, 0, 0};
 	return 0;
 }
 
 static void
 _moddeinit(void)
 {
-	chmode_table['O'].set_func = chm_nosuch;
-	chmode_table['O'].mode_type = 0;
+	chmode_table['O'] = (struct ChannelMode){chm_nosuch, 0, 0};
 }
 
 static void
 chm_operonly(struct Client *source_p, struct Channel *chptr,
-	int alevel, int parc, int *parn,
-	const char **parv, int *errors, int dir, char c, long mode_type)
+	int alevel, const char *arg, int *errors, int dir, char c, long mode_type)
 {
-	int newparn = 0;
-	const char *newparv[] = { "$o" };
-
 	if (MyClient(source_p)) {
-		chm_simple(source_p, chptr, alevel, parc, parn, parv,
+		chm_simple(source_p, chptr, alevel, NULL,
 				errors, dir, 'i', MODE_INVITEONLY);
-		chm_ban(source_p, chptr, alevel, 1, &newparn, newparv,
+		chm_ban(source_p, chptr, alevel, "$o",
 				errors, dir, 'I', CHFL_INVEX);
 	} else
-		chm_nosuch(source_p, chptr, alevel, parc, parn, parv,
+		chm_nosuch(source_p, chptr, alevel, NULL,
 				errors, dir, c, mode_type);
 }
--- a/extensions/chm_quietunreg_compat.c
+++ b/extensions/chm_quietunreg_compat.c
@@ -16,39 +16,31 @@ static const char chm_quietunreg_compat_desc[] =
 static int _modinit(void);
 static void _moddeinit(void);
 static void chm_quietunreg(struct Client *source_p, struct Channel *chptr,
-	int alevel, int parc, int *parn,
-	const char **parv, int *errors, int dir, char c, long mode_type);
+	int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
 
 DECLARE_MODULE_AV2(chm_quietunreg_compat, _modinit, _moddeinit, NULL, NULL, NULL, NULL, NULL, chm_quietunreg_compat_desc);
 
 static int
 _modinit(void)
 {
-	chmode_table['R'].set_func = chm_quietunreg;
-	chmode_table['R'].mode_type = 0;
-
+	chmode_table['R'] = (struct ChannelMode){ chm_quietunreg, 0, 0 };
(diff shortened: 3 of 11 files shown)


extensions/override.c
```diff
@@ -161,6 +161,12 @@ hack_channel_access(void *vdata)
 	if (data->approved == CHFL_CHANOP)
 		return;
 
+	if (data->dir == MODE_OP_QUERY)
+	{
+		sendto_one_notice(data->client, "Use operspy for privileged mode queries");
```
glguy, line 166: This is missing a leading `:`

glguy:
On seven when I run `MODE #glguy o glguy` that would mean "op glguy"

Now as an op I get a `401 glguy·*·No such nick/channel` error.

Without op (with and without override) I get `482 glguy·#glguy·You're not a channel operator`.

I miss the seven behavior that `+` was optional and the default and `-` switched over to delete. I don't know if other users or clients relied on the + being optional but it seems confusing that it won't be anymore.

It seems the the `+` continues to be optional for user modes. (yay)

glguy:
I don't seem to know how to use operspy to query an exempt list.

edk0:
> On seven when I run `MODE #glguy o glguy` that would mean "op glguy"
> 
> Now as an op I get a `401 glguy·*·No such nick/channel` error.
> 
> Without op (with and without override) I get `482 glguy·#glguy·You're not a channel operator`.

That would be a bug in this PR; it matches current behaviour, but I must have undone the change where I put the implicit + back. Should be fixed now.

> I don't seem to know how to use operspy to query an exempt list.

You can't; if we change override we'll need to add that. I haven't because I think it's worth talking about more: do we really override to change the +eI lists but not see them? Feels a bit weird. Maybe it's better to take the override change out of this PR and discuss it on its own.

glguy:
Minor style idea: you could save a bunch of duplication of the extern prototypes with this change.

```diff
diff --git a/include/channel.h b/include/channel.h
index bb1da83f..e666279a 100644
--- a/include/channel.h
+++ b/include/channel.h
@@ -121,7 +121,7 @@ struct ChModeChange
 	int mems;
 };

-typedef void (*ChannelModeFunc)(struct Client *source_p, struct Channel *chptr,
+/* some documentation about what the arguments mean */
+typedef void ChannelModeFunc(struct Client *source_p, struct Channel *chptr,
 		int alevel, const char *arg, int *errors, int dir, char c, long mode_type);

 enum chm_flags
@@ -136,7 +136,7 @@ enum chm_flags

 struct ChannelMode
 {
-	ChannelModeFunc set_func;
+	ChannelModeFunc *set_func;
 	long mode_type;
 	enum chm_flags flags;
 };
diff --git a/include/chmode.h b/include/chmode.h
index 850864a2..2ead3df9 100644
--- a/include/chmode.h
+++ b/include/chmode.h
@@ -34,30 +34,18 @@

 extern int chmode_flags[256];

-extern void chm_nosuch(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_orphaned(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_simple(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_ban(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_hidden(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_staff(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_forward(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_throttle(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_key(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_limit(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_op(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
-extern void chm_voice(struct Client *source_p, struct Channel *chptr,
-	   int alevel, const char *arg, int *errors, int dir, char c, long mode_type);
+extern ChannelModeFunc chm_nosuch;
+extern ChannelModeFunc chm_orphaned;
+extern ChannelModeFunc chm_simple;
+extern ChannelModeFunc chm_ban;
+extern ChannelModeFunc chm_hidden;
+extern ChannelModeFunc chm_staff;
+extern ChannelModeFunc chm_forward;
+extern ChannelModeFunc chm_throttle;
+extern ChannelModeFunc chm_key;
+extern ChannelModeFunc chm_limit;
+extern ChannelModeFunc chm_op;
+extern ChannelModeFunc chm_voice;

 extern unsigned int cflag_add(char c, ChannelModeFunc function);
 extern void cflag_orphan(char c);
diff --git a/ircd/chmode.c b/ircd/chmode.c
index 585c4363..8ed74ded 100644
--- a/ircd/chmode.c
+++ b/ircd/chmode.c
@@ -1512,7 +1512,7 @@ set_channel_mode(struct Client *client_p, struct Client *source_p,

 	for (ms = modesets; ms < mend; ms++)
 	{
-		ChannelModeFunc set_func = ms->cm->set_func;
+		ChannelModeFunc *set_func = ms->cm->set_func;
 		if (set_func == NULL)
 			set_func = chm_nosuch;
 		set_func(fakesource_p, chptr, alevel, ms->arg, &errors, ms->dir, ms->mode, ms->cm->mode_type);
```

glguy:
Do we have to have both `NULL` and `chm_nosuch`? It feels weird to keep checking for both.

glguy:
Maybe this is OK, but if I specify 5 mode changes 4 take effect (makes sense) but 5 get reported in the modehack snote. Maybe that's OK?

edk0:
I'm all on board with typedefing the function rather than the fucntion type; obviously I just changed one prototype and then vimmed, so I didn't really notice the extra work :)

I'd quite like to get rid of `chm_nosuch` too.

I'm also happy with the idea of doing limiting in the parser (I think it already has all the information it needs?) and therefore not reporting changes that hit limits in the canonicalised mode change string. But note that there will still be other cases where we report a change that didn't happen, like forwarding to a channel that doesn't exist, or banning an invalid extban.

Incidentally, let's remember to test whether solanum override lets you forward to a channel you aren't allowed to (it should).

glguy:
@edk0 Override does not appear to apply to forward

edk0:
I think I want to leave this PR where it is for now and make further changes in a new one. We've talked about lots of good things I'd like to do, but it's getting unwieldy, IMO.

glguy (approved):

PR #50, opened 2020-11-06
m_cap: simplify cap_req, remove multiline

dwfreed:
Spec doesn't really allow multiline here, so don't use it

### DIFF
--- a/modules/m_cap.c
+++ b/modules/m_cap.c
@@ -302,16 +302,11 @@ cap_ls(struct Client *source_p, const char *arg)
 static void
 cap_req(struct Client *source_p, const char *arg)
 {
-	static char buf_prefix[DATALEN + 1];
-	static char buf_list[2][DATALEN + 1];
-	const char *str_cont = " * :";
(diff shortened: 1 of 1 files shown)


modules/m_cap.c
```diff
@@ -302,16 +302,11 @@ cap_ls(struct Client *source_p, const char *arg)
 static void
 cap_req(struct Client *source_p, const char *arg)
 {
-	static char buf_prefix[DATALEN + 1];
-	static char buf_list[2][DATALEN + 1];
-	const char *str_cont = " * :";
-	const char *str_final = " :";
-	int len_prefix;
-	int max_list;
+	char ack_buf[BUFSIZE];
```
glguy, line 305: Let's make this `char ack_buf[DATALEN+1]` if that's what we need it to be.

dwfreed:
```suggestion
	char ack_buf[DATALEN+1];
```

glguy (approved):

PR #51, opened 2020-11-06
API for automatically splitting long messages in places where spec allows/requires it

### LINKED ISSUE #16: Centralise multiline reply generation

dwfreed:
This will greatly simplify and deduplicate code in various places, like CAP LS/LIST, NAMES, and WHOIS channel lists.

Fixes #16

### DIFF
--- a/include/send.h
+++ b/include/send.h
@@ -93,6 +93,18 @@ extern void kill_client(struct Client *client_p, struct Client *diedie,
 extern void kill_client_serv_butone(struct Client *one, struct Client *source_p,
 				    const char *pattern, ...) AFP(3, 4);
 
+enum multiline_item_result {
+	MULTILINE_FAILURE,
+	MULTILINE_SUCCESS,
+	MULTILINE_WRAPPED,
+};
+
+extern bool send_multiline_init(struct Client *target_p, const char *separator, const char *format, ...) AFP(3, 4);
+extern bool send_multiline_remote_pad(struct Client *target_p, struct Client *client_p);
+extern enum multiline_item_result send_multiline_item(struct Client *target_p, const char *format, ...) AFP(2, 3);
+extern bool send_multiline_fini(struct Client *target_p, const char *format, ...) AFP(2, 3);
+extern void send_multiline_reset(void);
+
 #define L_ALL 	0
 #define L_OPER 	1
 #define L_ADMIN	2
--- a/ircd/send.c
+++ b/ircd/send.c
@@ -1503,3 +1503,171 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char 
 
 	rb_linebuf_donebuf(&rb_linebuf_id);
 }
+
+static struct Client *multiline_stashed_target_p;
+static char multiline_prefix[DATALEN+1]; /* allow for null termination */
+static int multiline_prefix_len;
+static char multiline_separator[2];
+static int multiline_separator_len;
+static char *multiline_item_start;
+static char *multiline_cur;
+static int multiline_cur_len;
+static int multiline_remote_pad;
+
+bool
+send_multiline_init(struct Client *target_p, const char *separator, const char *format, ...)
+{
+	va_list args;
+
+	s_assert(multiline_stashed_target_p == NULL && "Multiline: didn't cleanup after last usage!");
+
+	va_start(args, format);
+	multiline_prefix_len = vsnprintf(multiline_prefix, sizeof multiline_prefix, format, args);
+	va_end(args);
+
+	if (multiline_prefix_len <= 0 || multiline_prefix_len >= DATALEN)
+	{
+		s_assert(false && "Multiline: failure preparing prefix!");
+		return false;
+	}
+
+	multiline_separator_len = rb_strlcpy(multiline_separator, separator, sizeof multiline_separator);
+	if (multiline_separator_len >= sizeof multiline_separator)
+	{
+		s_assert(false && "Multiline: separator too long");
+		return false;
+	}
+
+	multiline_stashed_target_p = target_p;
+	multiline_item_start = multiline_prefix + multiline_prefix_len;
+	multiline_cur = multiline_item_start;
+	multiline_cur_len = multiline_prefix_len;
+	multiline_remote_pad = 0;
+	return true;
+}
+
+bool
+send_multiline_remote_pad(struct Client *target_p, struct Client *client_p)
+{
+	ssize_t remote_pad;
+
+	if (target_p != multiline_stashed_target_p)
+	{
+		s_assert(false && "Multiline: missed init call!");
+		multiline_stashed_target_p = NULL;
+		return false;
+	}
+
+	if (MyConnect(target_p))
+		return true;
+
+	remote_pad = strlen(client_p->name) - strlen(client_p->id);
+
+	if (remote_pad > 0)
+	{
+		multiline_remote_pad += remote_pad;
+	}
+
+	return true;
+}
+
+enum multiline_item_result
+send_multiline_item(struct Client *target_p, const char *format, ...)
+{
+	va_list args;
+	char item[DATALEN];
+	int item_len, res;
+	enum multiline_item_result ret = MULTILINE_SUCCESS;
+
+	if (target_p != multiline_stashed_target_p)
+	{
+		s_assert(false && "Multiline: missed init call!");
+		multiline_stashed_target_p = NULL;
+		return MULTILINE_FAILURE;
+	}
+
+	va_start(args, format);
+	item_len = vsnprintf(item, sizeof item, format, args);
+	va_end(args);
+
+	if (item_len < 0 || multiline_prefix_len + multiline_remote_pad + item_len > DATALEN)
+	{
+		s_assert(false && "Multiline: failure preparing item!");
+		multiline_stashed_target_p = NULL;
+		return MULTILINE_FAILURE;
+	}
+
+	if (multiline_cur_len + ((*multiline_item_start != '\0') ? multiline_separator_len : 0) + item_len > DATALEN - multiline_remote_pad)
+	{
+		sendto_one(target_p, "%s", multiline_prefix);
+		*multiline_item_start = '\0';
+		multiline_cur_len = multiline_prefix_len;
+		multiline_cur = multiline_item_start;
+		ret = MULTILINE_WRAPPED;
+	}
+
+	res = snprintf(multiline_cur, sizeof multiline_prefix - multiline_cur_len, "%s%s",
+			(*multiline_item_start != '\0') ? multiline_separator : "",
+			item);
+
+	if (res < 0)
+	{
+		s_assert(false && "Multiline: failure appending item!");
+		multiline_stashed_target_p = NULL;
+		return MULTILINE_FAILURE;
+	}
+
+	multiline_cur_len += res;
+	multiline_cur += res;
+	return ret;
+}
+
+bool
+send_multiline_fini(struct Client *target_p, const char *format, ...)
+{
+	va_list args;
+	char final[DATALEN];
+	int final_len;
+
+	if (target_p != multiline_stashed_target_p)
+	{
+		s_assert(false && "Multiline: missed init call!");
+		multiline_stashed_target_p = NULL;
+		return false;
+	}
+
+	if (multiline_cur_len == multiline_prefix_len)
+	{
+		multiline_stashed_target_p = NULL;
+		return true;
+	}
+
+	if (format)
+	{
+		va_start(args, format);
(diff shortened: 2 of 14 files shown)


ircd/send.c
```diff
@@ -1503,3 +1503,139 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+		}
+	}
+	else
+	{
+		strncpy(final, multiline_prefix, sizeof final);
+		final_len = multiline_prefix_len;
+	}
+
+	/* strip trailing space */
+	*(multiline_cur - 1) = '\0';
```
edk0, line 1630:
```suggestion
	multiline_cur[-1] = '\0';
```

include/send.h
```diff
@@ -93,6 +93,17 @@ extern void kill_client(struct Client *client_p, struct Client *diedie,
 extern void kill_client_serv_butone(struct Client *one, struct Client *source_p,
 				    const char *pattern, ...) AFP(3, 4);
 
+enum multiline_item_result {
+	MULTILINE_FAILURE = 0,
+	MULTILINE_SUCCESS = 1,
+	MULTILINE_WRAPPED = 2,
```
edk0, line 99: No need to give explicit values here, IMO.

```suggestion
	MULTILINE_SUCCESS,
	MULTILINE_WRAPPED,
```

include/send.h
```diff
@@ -93,6 +93,17 @@ extern void kill_client(struct Client *client_p, struct Client *diedie,
...
+};
+
+extern bool sendto_one_multiline_init(struct Client *target_p, const char *format, ...) AFP(2, 3);
+extern enum multiline_item_result sendto_one_multiline_item(struct Client *target_p, const char *format, ...) AFP(2, 3);
+extern bool sendto_one_multiline_fini(struct Client *target_p, const char *format, ...) AFP(2, 3);
+extern void sendto_one_multiline_reset();
```
edk0, line 105:
```suggestion
extern void sendto_one_multiline_reset(void);
```

ircd/send.c
```diff
@@ -1503,3 +1503,139 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+	sendto_one(target_p, "%s%s", final, multiline_items);
+
+	multiline_stashed_target_p = NULL;
+	return true;
+}
+
+void
+sendto_one_multiline_reset()
```
edk0, line 1638:
```suggestion
sendto_one_multiline_reset(void)
```

ircd/send.c
```diff
@@ -1503,3 +1503,139 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+{
+	multiline_stashed_target_p = NULL;
```
edk0, line 1640: Should this complain if a multiline was not in progress?

edk0 (changes requested):
The missing `(void)` is the only serious problem here, and it's easy to address :)

One thing that crossed my mind and doesn't really fit anywhere specific in the diff is that it'd be pretty easy to keep the prefix in the same buffer we use to build items, but I don't see any great benefit beside the warm fuzzy feeling of saving a BUFSIZE.

ircd/send.c
```diff
@@ -1503,3 +1503,139 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+	{
+		sendto_realops_snomask(SNO_GENERAL, L_ALL, "Multiline: didn't cleanup after last usage!");
+		return false;
+	}
+
+	va_start(args, format);
+	multiline_prefix_len = vsnprintf(multiline_prefix, sizeof multiline_prefix, format, args);
+	va_end(args);
+
+	if (multiline_prefix_len < 0 || (size_t)multiline_prefix_len >= sizeof multiline_prefix - strlen("\r\n"))
```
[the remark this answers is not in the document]
dwfreed: Resolved by switching to `DATALEN` everywhere, which is `BUFSIZE - strlen("\r\n")`

ircd/send.c
```diff
@@ -1503,3 +1503,139 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+	int final_len;
+
+	if (target_p != multiline_stashed_target_p)
+	{
+		multiline_stashed_target_p = NULL;
+		sendto_realops_snomask(SNO_GENERAL, L_ALL, "Multiline: missed init call!");
+		return false;
+	}
+
+	if (multiline_cur_len == multiline_prefix_len)
```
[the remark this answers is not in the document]
dwfreed: (As noted on IRC) WHOIS channel list and NAMES may not produce any output, for example, but you don't know that until you've iterated through everything you might output.  This makes it so the consumer doesn't have to keep track of whether they produced any output before calling `_fini()`

ircd/send.c
dwfreed: I could see this go either way; so far the only usage is in CAP LS/LIST for < 3.2 clients, and not having this check for in-progress makes that implementation slightly simpler, because then it can just reset and return if result isn't success (no wrap), rather than reset only if wrapped, and then return if result isn't success.

modules/m_monitor.c
```diff
@@ -126,54 +100,60 @@ add_monitor(struct Client *client_p, const char *nicks)
...
-				*onptr++ = ',';
-				cur_onlen++;
-			}
+	RB_DLINK_FOREACH_PREV(ptr, prev_head->prev)
+	{
+		monptr = ptr->data;
 
-			arglen = sprintf(onptr, "%s!%s@%s",
-					target_p->name, target_p->username,
+		if ((target_p = find_named_person(monptr->name)) != NULL)
```
edk0, line 114: Common though this pattern is, I don't love it :(

ircd/send.c
```diff
@@ -1503,3 +1503,163 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+		if (final_len < 0 || final_len >= DATALEN || final_len > multiline_prefix_len)
+		{
+			s_assert(false && "Multiline: failure preparing final prefix!");
+			multiline_stashed_target_p = NULL;
+			return false;
+		}
+	}
+	else
+	{
+		rb_strlcpy(final, multiline_prefix, sizeof final);
```
edk0, line 1650: This adds an extra copy of the buffer before `multiline_item_start`.

```suggestion
		rb_strlcpy(final, multiline_prefix, multiline_prefix_len + 1);
```

ircd/send.c
```diff
@@ -1503,3 +1503,163 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+		s_assert(false && "Multiline: missed init call!");
+		multiline_stashed_target_p = NULL;
+		return MULTILINE_FAILURE;
+	}
+
+	va_start(args, format);
+	item_len = vsnprintf(item, sizeof item, format, args);
+	va_end(args);
+
+	if (item_len < 0 || item_len >= DATALEN || multiline_prefix_len + multiline_remote_extra_space + item_len > DATALEN)
```
edk0, line 1584: Does `item_len >= DATALEN` need to be here?

ircd/send.c
```diff
@@ -1503,3 +1503,163 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+
+	if (res < 0)
+	{
+		s_assert(false && "Multiline: failure appending item!");
+		multiline_stashed_target_p = NULL;
+		return MULTILINE_FAILURE;
+	}
+
+	multiline_cur_len += res;
+	multiline_cur += res;
```
edk0, line 1611: This would have potential undefined behaviour if the buffer didn't contain headroom for the longest allowable separator. I *think* if we did this (and the trailing separator skip above and in `sendto_one_multiline_fini`) some other way we wouldn't need that headroom.

ircd/send.c
```diff
@@ -1503,3 +1503,163 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+	va_end(args);
+
+	if (multiline_prefix_len < 0 || multiline_prefix_len >= DATALEN)
+	{
+		s_assert(false && "Multiline: failure preparing prefix!");
+		return false;
+	}
+
+	multiline_stashed_target_p = target_p;
+	multiline_separator_len = rb_strlcpy(multiline_separator, separator, sizeof multiline_separator);
```
edk0, line 1535:
```suggestion
	multiline_separator_len = rb_strlcpy(multiline_separator, separator, sizeof multiline_separator);
	if (multiline_separator_len >= sizeof multiline_separator)
	{
	    s_assert(false && "Multiline: separator too long");
	    return false;
	}
```

ircd/send.c
```diff
@@ -1503,3 +1503,163 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+		return true;
+	}
+
+	if (format)
+	{
+		va_start(args, format);
+		final_len = vsnprintf(final, sizeof final, format, args);
+		va_end(args);
+
+		if (final_len < 0 || final_len >= DATALEN || final_len > multiline_prefix_len)
```
edk0, line 1641: Do we need `final_len >= DATALEN`?

edk0 (changes requested):
I made a start on a test suite for this while reviewing it: https://github.com/solanum-ircd/solanum/commit/56890e28a6892de0d52de0e14a8a007575edf74d—feel free to pull it into the PR if you think it's useful.

ircd/send.c
dwfreed: We could do this the other way, put the separator first then the item, and only use the separator if `*multiline_item_start != '\0'`.  Then it wouldn't be necessary to include the separator space in the size of the buffer

ircd/send.c
dwfreed: This was fine until *you* wanted to eliminate the separate items buffer 😛

ircd/send.c
edk0: We don't need to include the separator space either way; if it fills the buffer there'll be no room for more items, so we're not going to send it. We'd just need to cut it off with a null character in a way that doesn't depend on having enough room to sprintf all of it into the buffer.

ircd/send.c
```diff
@@ -1503,3 +1503,169 @@ kill_client_serv_butone(struct Client *one, struct Client *target_p, const char
...
+{
+	ssize_t remote_extra_space;
+
+	if (target_p != multiline_stashed_target_p)
+	{
+		s_assert(false && "Multiline: missed init call!");
+		multiline_stashed_target_p = NULL;
+		return false;
+	}
+
```
edk0, line 1560: Why not check MyConnect here and save callers doing it?

```suggestion

	if (!MyConnect(client_p))
		return true;

```

modules/m_cap.c
```diff
@@ -152,31 +152,29 @@ clicap_find(const char *data, int *negate, int *finished)
...
 			EmptyString(source_p->name) ? "*" : source_p->name,
-			subcmd);
+			subcmd,
+			(source_p->flags & FLAGS_CLICAP_DATA) ? str_cont : str_final);
 
 	/* shortcut, nothing to do */
-	if (flags == -1 || len_prefix < 0) {
-		sendto_one(source_p, "%s%s", buf_prefix, str_final);
+	if (flags == -1 || !multiline_ret) {
+		sendto_one(source_p, "%s CAP %s %s %s",
```
edk0, line 170:
```suggestion
		sendto_one(source_p, ":%s CAP %s %s %s",
```

edk0 (changes requested):
Modulo suggestions I think this is in a pretty good place now.

I think it might be worth raising, since this will be the last good opportunity to filter-branch, the question of whether we could bikeshed the names down a bit. `sendto_one_multiline_remote_extra_space` doesn't exactly roll off the tongue.

modules/m_monitor.c
```diff
@@ -126,54 +100,61 @@ add_monitor(struct Client *client_p, const char *nicks)
...
 
-	if(cur_onlen != mlen)
-		sendto_one(client_p, "%s", onbuf);
-	if(cur_offlen != mlen)
-		sendto_one(client_p, "%s", offbuf);
+	sendto_one_multiline_fini(client_p, NULL);
+
+	if (name)
+	{
+		char buf[DATALEN];
```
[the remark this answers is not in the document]
dwfreed: I disagree with leaving this at 100, as it would significantly truncate the input if it's long (6 16 character nicks would be the point at which it would start truncating, which is why I increased it in the first place).  An estimated calculation of what the rest of the numeric would need for space leaves a bit more than 400 for this parameter (NICKLEN 32 * 2 + list limit 3 + numeric 3 + spaces and leading colon 6 + trailing string 21 = 97; 510 - 97 = 413), so that's what I'll use here

edk0 (approved):
🍰

PR #53, opened 2020-11-08
Stop using chm_nosuch as a sentinel value

glguy:


edk0 (dismissed):
It'd be nice to get rid of `chm_nosuch` altogether, given what it's still doing.

Maybe the `_compat` modules could disable themselves if the underlying modes they depend on aren't enabled?

edk0 (approved):

PR #56, opened 2020-11-08
Clean up things related to building tests

edk0:
Now adding a new test program should be a matter of adding it to `check_PROGRAMS`; the other 3 places you had to mess with before will look after themselves.


glguy:
If `TESTS` is generated now, should `make clean` delete it?

dwfreed (approved):

PR #58, opened 2020-11-08
Add general::stats_l_oper_only

edk0:


ircd/newconf.c
```diff
@@ -1542,6 +1542,21 @@ conf_set_general_stats_i_oper_only(void *data)
...
+	char *val = data;
+
+	if(rb_strcasecmp(val, "yes") == 0)
+		ConfigFileEntry.stats_l_oper_only = 2;
+	else if(rb_strcasecmp(val, "self") == 0)
+		ConfigFileEntry.stats_l_oper_only = 1;
+	else if(rb_strcasecmp(val, "no") == 0)
+		ConfigFileEntry.stats_l_oper_only = 0;
+	else
+		conf_report_error("Invalid setting '%s' for general::stats_l_oper_only.", val);
```
glguy, line 1557: I'd prefer if the error messages also listed what the valid values are when the set of them is small, but I looked around and the rest of the configuration processing behaves this same way, so that's something we could worry about separately.

include/s_conf.h
```diff
@@ -191,6 +191,7 @@ struct config_file_entry
 	int stats_h_oper_only;
 	int stats_o_oper_only;
 	int stats_k_oper_only;
+	int stats_l_oper_only;
```
glguy, line 194: An enum seems nicer than unnamed magic int values.

glguy (dismissed):
I have some comments for possible improvement, but the changes themselves look good.

modules/m_stats.c
```diff
@@ -1523,12 +1523,14 @@ stats_ltrace(struct Client *source_p, int parc, const char *parv[])
 	const char *name;
 	char statchar = parv[1][0];
 
+	if (ConfigFileEntry.stats_l_oper_only == 2 && !IsOperGeneral(source_p))
+		sendto_one_numeric(source_p, ERR_NOPRIVILEGES, form_str(ERR_NOPRIVILEGES));
```
glguy, line 1527: Without a `return` this privilege check doesn't appear to enforce anything.

```suggestion
	if (ConfigFileEntry.stats_l_oper_only == 2 && !IsOperGeneral(source_p)) {
		sendto_one_numeric(source_p, ERR_NOPRIVILEGES, form_str(ERR_NOPRIVILEGES));
		return;
	}
```

glguy (changes requested):

modules/m_stats.c
edk0: oops :)

glguy: alternatively the change might be a following `else if`

glguy:
We'll need to add this setting to `INFO`'s operator output.

glguy (approved):

PR #63, opened 2020-11-08
Add /stats s to list secure blocks

edk0:
`s` and `S` were both used for ssld stats, but only `S` was documented, so I'm taking `s`. You can't stop me!


glguy (approved):
Test it and got the expected output. Privilege check worked.

PR #73, opened 2020-11-14
Get rid of hub_mask/leaf_mask

edk0:


glguy:
- `doc/oper-guide/config.rst` and `doc/oper-guide/commands.rst` talk about hubs and leaves.
- `doc/technical/capab.txt` mentions a HUB capability, does this still exist?

edk0:
I don't think it does. About the only thing in doc/ that's been kept (mostly) up-to-date is the ts6-protocol.txt. I don't mind fixing this, but I think we should either decide we're going to bring them up to date in general, or decide we can't be bothered and mark them as historical interest.

glguy (approved):
Code changes look lgtm

PR #89, opened 2020-12-12
Use Github Actions

dwfreed:
One possible addition to this would be a problem matcher that will automatically annotate lines with compiler output.


dwfreed:
Travis will never finish because I killed the .travis.yml in this PR.

xfnw (approved):
huh weird, the github actions passing badge is a slightly minty-er shade of green than the old travis one

looks good though

swantzter:
I noticed we generate html for the oper guide (and I've had the thought in the back of my head for a bit anyways) maybe it's worth considering building and publishing the oper guide to the site? (it'd be a separate issue ofc)

edk0:
> I noticed we generate html for the oper guide (and I've had the thought in the back of my head for a bit anyways) maybe it's worth considering building and publishing the oper guide to the site? (it'd be a separate issue ofc)

Not until it's updated

swantzter:
> Not until it's updated

Obviously, meant it more as a "sometimes in the future" thing

edk0 (approved):

PR #93, opened 2020-12-20
Make most snotes NETWIDE

dwfreed:
I did look through most of these to make sure they were sane, but a second set of eyes would be good.  I didn't touch most snotes for bans (K/D/X lines, RESVs) because I figured we could do those when we rework ban propagation and elimination of clusters and such.


edk0:
I'm not thrilled about making things netwide snotes that are propagated globally anyway.

dwfreed:
Connect needs to be netwide, because only the local server knows the connection class; the others that are propagated that previously had a snote extension I'd be open to integrating into the appropriate place.  But sno_globaloper was a hack (checks for umode +o), and sno_globalkline is the hackiest of hacks (checks for specific quit messages), so those definitely need to go.

dwfreed:
(The sasl_abort1 test failures are because the connect snotes are now in the server sendq)

PR #94, opened 2020-12-24
chmode: centralise + test mode limits

edk0:
One tangible effect of this is that mode changes in oper overrides will be more accurate.


tests/chmode1.c
```diff
@@ -66,6 +66,41 @@ test_chmode_parse(void)
 	remove_hook("get_channel_access", chmode_access_hook);
 }
 
+void
+test_chmode_limits(void)
+{
+	char chmode_buf[2 + MAXMODEPARAMS + 1] = "+";
```
dwfreed, line 72: Why is this 2 + ... instead of 1 + ... ?

dwfreed (commented):
Just one question that may be due to me not being able to brain

tests/chmode1.c
edk0: The string is `+{'l' * (MAXMODEPARAMS+1)}\0` and I thought `1 + 1` looked silly

dwfreed: Aha, missed the for loop after was using `MAXMODEPARAMS + 1`

dwfreed (approved):

PR #98, opened 2021-01-01
send pseudo MODE for chm_hidden modes on oper_up

jesopo:
placement of this code might be wrong. it's quadratic which I think is probably fine unless we have loads of `chm_hidden` modes or so. usual caveat that I'm not great at C

### DIFF
--- a/ircd/s_user.c
+++ b/ircd/s_user.c
@@ -1400,7 +1400,12 @@ user_welcome(struct Client *source_p)
 void
 oper_up(struct Client *source_p, struct oper_conf *oper_p)
 {
-	unsigned int old = source_p->umodes, oldsnomask = source_p->snomask;
+	unsigned int old = source_p->umodes, oldsnomask = source_p->snomask, i = 0;
+	rb_dlink_node *ptr;
+	struct Channel *chptr;
+	char cmode1[256], cmode2[256];
+	char *cmode_ptr = cmode1;
+
 	hook_data_umode_changed hdata;
 
 	SetOper(source_p);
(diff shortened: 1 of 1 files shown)


ircd/s_user.c
```diff
@@ -1400,7 +1400,12 @@ user_welcome(struct Client *source_p)
 void
 oper_up(struct Client *source_p, struct oper_conf *oper_p)
 {
-	unsigned int old = source_p->umodes, oldsnomask = source_p->snomask;
+	unsigned int old = source_p->umodes, oldsnomask = source_p->snomask, i = 0;
```
edk0, line 1403: Style: use a new declaration for unrelated variables

ircd/s_user.c
```diff
@@ -1400,7 +1400,12 @@ user_welcome(struct Client *source_p)
...
+	rb_dlink_node *ptr;
+	struct Channel *chptr;
+	unsigned char cmode1[256], cmode2[256];
+	unsigned char *cmode_ptr = cmode1;
```
edk0, line 1407: Style: could these have better names?

ircd/s_user.c
```diff
@@ -1469,6 +1474,29 @@ oper_up(struct Client *source_p, struct oper_conf *oper_p)
...
+	if (HasPrivilege(source_p, "auspex:cmodes"))
+	{
+		for (i = 0; i < 256; i++)
+			if (chmode_table[i].set_func == chm_hidden)
+				*cmode_ptr++ = i;
+		*cmode_ptr = '\0';
+
+		RB_DLINK_FOREACH(ptr, source_p->user->channel.head)
+		{
+			chptr = ((struct membership *)ptr->data)->chptr;
```
edk0, line 1487: Style: avoid casts

ircd/s_user.c
```diff
@@ -1469,6 +1474,29 @@ oper_up(struct Client *source_p, struct oper_conf *oper_p)
...
+			cmode_ptr = cmode2;
+
+			for (i = 0; cmode1[i]; i++)
+				if (chptr->mode.mode & chmode_flags[cmode1[i]])
+					*cmode_ptr++ = cmode1[i];
+			*cmode_ptr = '\0';
+
+			if (*cmode2)
```
edk0, line 1495:
```suggestion
			if (cmode2[0] != '\0')
```

ircd/s_user.c
```diff
@@ -1469,6 +1474,29 @@ oper_up(struct Client *source_p, struct oper_conf *oper_p)
 	send_umode_out(source_p, source_p, old);
 	sendto_one_numeric(source_p, RPL_SNOMASK, form_str(RPL_SNOMASK),
 		   construct_snobuf(source_p->snomask));
+
+	if (HasPrivilege(source_p, "auspex:cmodes"))
```
edk0, line 1478: You don't `oper_up` when your privset changes, so I think this is the wrong place to do anything that checks privs. Currently the best place is probably `user_mode`, hacky though it is, since we arrange to call it every time privs change.

edk0 (changes requested):
I think this is a good change. I've got some nitpicks (and one important thing) in the review comments. I also wonder if it's worth sending RPL_CHANNELMODEIS rather than MODE if it'll do the same job of keeping clients in sync?

ircd/s_user.c
jesopo: doesn't it also get called for other stuff? how are we meant to know when `/mode mynick` isn't supposed to send out these `MODE`s?

edk0: I don't think we can :(

Maybe it's time to bite the bullet and add a standard mechanism for all privilege changes

ircd/s_user.c
jesopo: new names any better?

ircd/s_user.c
jesopo: what does this involve

jesopo:
do we also want to send the removal of these modes if someone either deopers are drops to a privset without `auspex:cmodes`?

edk0:
I think that'd be ideal, yeah.

jesopo:
then when we add a "standard mechanism" for priv changes, we'd want to make sure it's also called for changing to no privs

jesopo:
depends on #112

PR #102, opened 2021-01-22
Add +R channel mode module requiring services account to chat

glguy:
I'm not invested in any of the naming, reply numbers, error text, etc. If different words are in order let's have it!

I used error 415 to match OFTC's behavior with it's own +M channel mode.


modules/chm_regmsg.c
```diff
@@ -0,0 +1,88 @@
...
+#include "numeric.h"
+#include "chmode.h"
+#include "inline/stringops.h"
+
+static const char chm_regmsg_desc[] =
+	"Adds channel mode +R, which blocks messagess from unregistered users";
+
+static unsigned int mode_regmsg;
+
+static void chm_regmsg_process(hook_data_privmsg_channel *);
```
edk0, line 42: I know there's still some more of this in the code, but can we avoid the UB in favour of converting from `void *` inside the function?

modules/chm_regmsg.c
```diff
@@ -0,0 +1,88 @@
...
+	/* message is already blocked, defer */
+	if (data->approved)
+		return;
+
+	/* user is identified, accept */
+	if (!EmptyString(data->source_p->user->suser))
+		return;
+
+	/* voice and op override identification requirement, accept */
+	msptr = find_channel_membership(data->chptr, data->source_p);
```
edk0, line 63: It's pretty wack that `hook_data_privmsg_channel` doesn't have a membership on it, but I'm not sure it needs to be fixed here.

modules/chm_regmsg.c
```diff
@@ -0,0 +1,88 @@
...
+	sendto_one_numeric(data->source_p, ERR_MSGNEEDREGGEDNICK, form_str(ERR_MSGNEEDREGGEDNICK),
+		data->source_p->name, data->chptr->chname);
+	data->approved = ERR_MSGNEEDREGGEDNICK;
+}
+
+static int
+_modinit(void)
+{
+	mode_regmsg = cflag_add('R', chm_simple);
+	if (mode_regmsg == 0)
```
edk0, line 76: Probably worth spewing some sort of error message here.

edk0 (changes requested):

modules/chm_regmsg.c
glguy: If we added it, I'd want the field not to be pre-computed by simply a cache and we could have a helper function that either returns the computed value or computes it and then caches it. I think we don't actually need membership for the other users of this hook.

modules/chm_regmsg.c
glguy: spewed

modules/chm_regmsg.c
edk0: It's almost always going to be computed anyway in `can_send`

edk0 (approved):

PR #109, opened 2021-01-27
change "forcing message" override from CAN_SEND_NONOP to CAN_SEND_OPV

### LINKED ISSUE #108: make overriding users immune to channel throttles

jesopo:
per
https://github.com/solanum-ircd/solanum/blob/a92275551218402bf4969a7fe613cab27416c4ef/modules/core/m_message.c#L550-L551
we want to avoid the `flood_attack_channel` call, and I can't see `CAN_SEND_OPV` being used anywhere else for anything more serious than this

closes #108

### DIFF
--- a/extensions/override.c
+++ b/extensions/override.c
@@ -217,12 +217,12 @@ hack_can_send(void *vdata)
 	if (data->dir == MODE_QUERY)
 		return;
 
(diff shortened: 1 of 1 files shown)


jesopo:
this would mean you're overriding every message

PR #112, opened 2021-01-29
Track privset changes

### LINKED ISSUE #107: give modules a way to see which privs changed when privs change

edk0:
Make privsets a little heavier so it's easy to diff them, do that, and fire a hook to inform modules of the changes whenever someone's privset changes. Also add tests.

closes #107

### DIFF
--- a/ircd/privilege.c
+++ b/ircd/privilege.c
@@ -319,7 +319,7 @@ privilegeset_unref(struct PrivilegeSet *set)
 	}
 }
 
-const struct PrivilegeSet **
+struct privset_diff
 privilegeset_diff(const struct PrivilegeSet *old, const struct PrivilegeSet *new)
 {
 	static const char *no_privs[] = { NULL };
@@ -327,16 +327,15 @@ privilegeset_diff(const struct PrivilegeSet *old, const struct PrivilegeSet *new
 	static struct PrivilegeSet *set_unchanged = NULL,
 	                           *set_added = NULL,
 	                           *set_removed = NULL;
-	static const struct PrivilegeSet *result_sets[3];
 	static size_t n_privs = 0;
 	size_t new_size = n_privs ? n_privs : 32;
 	size_t i = 0, j = 0;
 
-	if (result_sets[0] == NULL)
+	if (set_unchanged == NULL)
 	{
-		result_sets[0] = set_unchanged = privilegeset_new_orphan("<unchanged>");
-		result_sets[1] = set_added = privilegeset_new_orphan("<added>");
-		result_sets[2] = set_removed = privilegeset_new_orphan("<removed>");
+		set_unchanged = privilegeset_new_orphan("<unchanged>");
+		set_added = privilegeset_new_orphan("<added>");
+		set_removed = privilegeset_new_orphan("<removed>");
 	}
 
 	if (old == NULL)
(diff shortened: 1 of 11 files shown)


ircd/privilege.c
```diff
@@ -58,6 +66,178 @@ privilegeset_get_any(const char *name)
...
+}
+
+static void
+privilegeset_index(struct PrivilegeSet *set)
+{
+	size_t n;
+	const char *s;
+	const char **p;
+
+	if (set->privs != NULL)
```
glguy, line 83: rb_free already checks `NULL`

ircd/privilege.c
```diff
@@ -58,6 +66,178 @@ privilegeset_get_any(const char *name)
...
+	set->privs = NULL;
+	set->priv_storage = NULL;
+	set->status = 0;
+	set->refs = 0;
+	set->name = rb_strdup(name);
+	return set;
+}
+
+static void
+privilegeset_free(struct PrivilegeSet *set)
```
glguy, line 186: it's nice when *free* functions tolerate `NULL`. Instead of testing `set -> shadow` for `NULL`, let's test `set` for `NULL` and then we can unconditionally free `->shadow` like we unconditionally free everything else.

ircd/privilege.c
```diff
@@ -156,40 +326,118 @@ privilegeset_unref(struct PrivilegeSet *set)
...
+
+	const char **res_unchanged = set_unchanged->privs;
+	const char **res_added = set_added->privs;
+	const char **res_removed = set_removed->privs;
+
+	while (i < old->size || j < new->size)
+	{
+		const char *oldpriv = NULL, *newpriv = NULL;
+		int ord = 0;
+		if (i < old->size)
```
glguy, line 376: We shouldn't need these conditionals. The priv lists are NULL terminated, so i and j can just get stuck pointing to the NULL at the end.

ircd/s_user.c
```diff
@@ -972,6 +972,33 @@ report_and_set_user_flags(struct Client *source_p, struct ConfItem *aconf)
 	}
 }
 
+void
+report_priv_change(struct Client *client, struct PrivilegeSet *old, struct PrivilegeSet *new)
+{
+	const struct PrivilegeSet *added, *removed, *unchanged;
+	if (old == NULL && new == NULL)
```
glguy, line 979: Should this logic be in s_conf.c and then that file can call into this function with the right old and new? Do we expect this behavior anywhere else?

ircd/s_user.c
```diff
@@ -972,6 +972,33 @@ report_and_set_user_flags(struct Client *source_p, struct ConfItem *aconf)
...
+	{
+		new = client->user->privset;
+		if (new != NULL)
+			old = new->shadow;
+	}
+
+	const struct PrivilegeSet **result = privilegeset_diff(old, new);
```
glguy, line 986: I think these three results should get freed at some point. They are allocated with new_orphan which mallocs

edk0: I think I'd prefer not to; currently they're allocated once and then reused for the life of the program, so I think of them as morally static objects that happen to be allocated in practice.

glguy: I missed that there was a test such that these are allocated only once

glguy (approved):
I had a couple of comments, but the change seems good! For things like initialization I think we're getting a pass from rb_alloc zeroing out memory before returning it

glguy (dismissed):

glguy (approved):

PR #116, opened 2021-02-01
OpenSSL: Support configuration of TLSv1.3 ciphersuites

aaronmdjones:
The OpenSSL developers decided, during the OpenSSL 1.1.1 development phase, to use a different API and different set of lists for TLSv1.3 ciphersuites, than for every TLS version preceeding it.

This is stupid, but we have to work with it.

This commit also improves configuration fault resilience. The reason is that if you don't pass any valid old-style ciphersuites, OpenSSL will not negotiate an older protocol at all. However, when they implemented the new API, they decided that lack of any valid ciphersuites should result in using the defaults. This means that if you pass a completely invalid ciphersuite list (like "foo"), OR if you pass a TLSv1.2-only ciphersuite list, TLSv1.3 continues to work. This is not mirrored; passing a TLSv1.3-only ciphersuite list will break TLSv1.2 and below.

Therefore we work around this lack of mirroring by falling back to the default list for each protocol. This means that if `ssl_cipher_list` is complete garbage, the default will be used, and TLS setup will succeed for both protocols. This is logged, so that administrators can fix their configuration.

I prefer this approach over explicitly disabling the protocols if their respective ciphersuite lists are invalid, because it will result in unusable TLSv1.3 if people run newer solanum with their older charybdis/solanum configuration files that contain custom `ssl_cipher_list` definitions. Hindering TLSv1.3 adoption is not an option, in my opinion.

The downside of this is that it is no longer possible to disable a protocol family by not including any of its ciphersuites. This could be remedied by an `ssl_protocol_list` configuration directive if it is decided that this functionality is ultimately necessary.

This work is not required for either of the other TLS backends, because neither of those libraries yet support TLSv1.3, and in the event that they eventually do, I expect them to allow configuration of newer ciphersuites with the existing APIs. This can be revisited if it turns out not to be the case.

### DIFF
--- a/librb/src/openssl.c
+++ b/librb/src/openssl.c
@@ -413,12 +413,40 @@ rb_setup_ssl_server(const char *const certfile, const char *keyfile,
 		}
 	}
 
-	if(SSL_CTX_set_cipher_list(ssl_ctx_new, cipherlist) != 1)
+
+	int ret_old = SSL_CTX_set_cipher_list(ssl_ctx_new, cipherlist);
+
+	if (ret_old != 1)
 	{
-		rb_lib_log("%s: SSL_CTX_set_cipher_list: could not configure any ciphers", __func__);
+		(void) rb_lib_log("%s: no valid old-style ciphersuites found "
+		                  "in ssl_cipher_list; will use defaults",
+		                  __func__);
+
+		ret_old = SSL_CTX_set_cipher_list(ssl_ctx_new, rb_default_ciphers);
+	}
+
+	#ifndef LRB_HAVE_TLS13
+	int ret_new = 0;
+	#else
+	int ret_new = SSL_CTX_set_ciphersuites(ssl_ctx_new, cipherlist);
+
+	if (ret_new != 1)
+	{
+		(void) rb_lib_log("%s: no valid new-style ciphersuites found "
+		                  "in ssl_cipher_list; will use defaults",
+		                  __func__);
+
+		ret_new = SSL_CTX_set_ciphersuites(ssl_ctx_new, rb_default_ciphers);
+	}
+	#endif
+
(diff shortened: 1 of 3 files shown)


librb/src/openssl.c
```diff
@@ -413,13 +423,41 @@ rb_setup_ssl_server(const char *const certfile, const char *keyfile,
 		}
 	}
 
-	if(SSL_CTX_set_cipher_list(ssl_ctx_new, cipherlist) != 1)
+
+	int ret1 = SSL_CTX_set_cipher_list(ssl_ctx_new, cipherlist_old);
```
edk0, line 427: `ret_old` and `ret_new` might make this bit easier to follow

librb/src/openssl.c
```diff
@@ -413,13 +423,41 @@ rb_setup_ssl_server(const char *const certfile, const char *keyfile,
...
+		                  "in ssl_cipher_list; will use defaults",
+		                  __func__);
+
+		ret2 = SSL_CTX_set_ciphersuites(ssl_ctx_new, rb_default_ciphers_new);
+	}
+	#endif
+
+	if (ret1 != 1 && ret2 != 1)
+	{
+		rb_lib_log("%s: could not configure any ciphers", __func__);
```
edk0, line 455: Are you able to hit this in tests? Should it say which of old/new failed?

It looks to me like this path is inevitable if !defined(LRB_HAVE_TLS13)

aaronmdjones: The if tests above it already say which of old and/or new failed.

librb/src/openssl.c
aaronmdjones: Done.

edk0 (approved):

PR #117, opened 2021-02-01
Extend oper:override for +f oper requirement

supertassu:


edk0:
Thanks for working on that issue! I think we'd ideally like to solve this more generally, by making the core mechanisms here more extensible so that the logic of this change can go in the override extension.

I think that'd boil down to calling `get_channel_access` unconditionally everywhere it's currently gated on membership and, in `check_forward` specifically, calling it with a "direction" other than `MODE_QUERY`, as that's (ab)used to signal that we're not making changes. If you don't do anything else, extensions/override will already implement the desired behaviour, as well as allow overriding anything else from outside channels. I think that's fine, but it might merit wider discussion.

What that won't do is report setting +f where the overriding oper has ops in the source channel but not the target, and I think that's worth having. For that we'd need to create a new hook (along the lines of can_kick). We could call that from check_forward and hook it in extensions/override to emit the usual override spam; again copying how it deals with kicks seems a good starting point.

I'd be happy to step through things over IRC if you're not sure about any of this.

PR #135, opened 2021-04-06
Show account name in cliconn snotes when SASL is used

examknow:
Per #91

Using SASL: `*** Notice -- Client connecting: nick (user@host) [1.2.3.4] {users} [gecos] <AccountName>`
Without SASL: `*** Notice -- Client connecting: nick (user@host) [1.2.3.4] {users} [gecos] <*>`


glguy:
I think it might be worth putting the account name just before the gecos field as the gecos field just to make it a bit easier to match theses notes correctly given the very permissive nature of that field. `]$` is currently a reliable way to match that the end of that field.

(this isn't an instruction, I'm interested in what others think)

dwfreed:
I would like to see this applied to the machine-parsable cliconn snotes as well (with the ordering @glguy suggested, account then gecos)

jesopo:
machine readable cliconns being `+C`?

examknow:
Requested changes made in the above commit.

glguy (approved):

Issue #85, opened 2020-11-30
support more human friendly k/d/x-line duration format


jesopo:
example:
`KLINE 11520 *@1.2.3.4 :go away`
could instead be:
`KLINE 1w1d *@1.2.3.4 :go away`

I think we'd want to support **w**eek, **d**ay, **h**our, **m**inute

I think this is where would want changing....
https://github.com/solanum-ircd/solanum/blob/6485005214c12ef94a50096e96478804b9cef556/ircd/s_newconf.c#L684-L685

glguy:
This feels like the kind of thing people should be programming their clients to handle nicely; people shouldn't be manually quoting KLINE commands anyway.

The KLINE format is already a mess with optional duration and server parameters spread around.

I guess I buy the argument that this change doesn't hurt, so why not, but it feels like it's in the wrong place to me.

dwfreed:
Fixed by #317

PR #136, opened 2021-04-07
Support more human friendly k/d/x-line duration format

examknow:
Per #85

K/D/X-Line durations can now be set like `KLINE 3w2d1m *@1.2.3.4 :test` and have the same effect as `KLINE 33121 *@1.2.3.4 :test`. K-Lines set using the old duration format will still work the same as before.


ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
+	while (*ch) {
+		och = ch;
+		current = strtol(ch, &ch, 10);
+		if (och == ch)
+			/* No numbers were given so return invalid */
 			return -1;
+		if (current) {
+			switch (*ch++) {
+				case '\0':
+					/* No unit was given so send it back as seconds */
```
edk0, line 700: comment says seconds, we interpret it as minutes. confusing?

ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
+					break;
+				default:
+					/* Bad time unit was given */
+					return -1;
+			}
+		}
 	}
-
-	if(result > (60 * 24 * 7 * 52))
-		result = (60 * 24 * 7 * 52);
```
edk0, line 702: I think removing this check is out of scope.

ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
-	{
-		if(IsDigit(*p))
-		{
-			result *= 10;
-			result += ((*p) & 0xF);
-			p++;
-		}
-		else
+	int current = 0;
+	char *ch = (char *)p, *och;
```
edk0, line 689: I know it'll make the strtol more cumbersome, but I don't think that's a good enough reason to cast.

ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
-	while(*p)
-	{
-		if(IsDigit(*p))
-		{
-			result *= 10;
-			result += ((*p) & 0xF);
-			p++;
-		}
-		else
+	int current = 0;
```
edk0, line 688: Wouldn't long be better, since that's what strtol returns? (Or unsigned long and strtoul?)

ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
+	char *ch = (char *)p, *och;
+
+	while (*ch) {
+		och = ch;
+		current = strtol(ch, &ch, 10);
```
edk0, line 693: Please check for range errors (return value of LONG_MIN/LONG_MAX + errno == ERANGE)

ircd/s_newconf.c
```diff
@@ -685,23 +685,39 @@ time_t
...
+					break;
+				case 'h':
+					result += (current * 3600);
+					break;
+				case 'd':
+					result += (current * 86400);
+					break;
+				case 'w':
+					result += (current * 604800);
+					break;
```
edk0, line 713: It would make sense to me to check for overflow for these

edk0 (changes requested):
Thanks for the PR! 🍰

ircd/s_newconf.c
```diff
@@ -685,23 +685,53 @@ time_t
...
+				case 'w':
+					if (errno == ERANGE)
+						return -1;
+					result += (current * 604800);
+					break;
+				default:
+					/* Bad time unit was given */
+					return -1;
+			}
```
dwfreed, line 729: All these ERANGE tests are redundant with the one from above.

The null handler will bypass the cap on max duration below.

After fixing the above bypass, this will result in an out of bounds read in the while test if no unit is given for the last number

```suggestion
			switch (*ch) {
				case '\0': /* No unit was given so send it back as minutes */
				case 'm':
					result += (current * 60);
					break;
				case 'h':
					result += (current * 3600);
					break;
				case 'd':
					result += (current * 86400);
					break;
				case 'w':
					result += (current * 604800);
					break;
				default:
					/* Bad time unit was given */
					return -1;
			}
			if (*ch++ == '\0')
				break;
```

ircd/s_newconf.c
```diff
@@ -685,23 +685,53 @@ time_t
...
+		}
 	}
-
 	if(result > (60 * 24 * 7 * 52))
 		result = (60 * 24 * 7 * 52);
```
dwfreed, line 733: Because you're doing the minute unit multiplication before here, you need to include it in the cap test.

```suggestion
	if(result > (60 * 60 * 24 * 7 * 52))
		result = (60 * 60 * 24 * 7 * 52);
```

dwfreed (changes requested):

jesopo (approved):
wfm, lgtm

edk0:
weird style (`result += (current * 60);`?) aside, this still has a memory leak

dwfreed:
Superseded by #317

Issue #122, opened 2021-02-07
Support shedding users


dwfreed:
Shedding actively disconnects users from the ircd, so that they can reconnect to the round-robin.  It does this one user at a time, at an oper-specified interval (with random variation added).  It's usually configured to exclude opers, but not always.  The syntax from OFTC help:

```
SHEDDING <server> <approx_seconds_per_userdrop> <opers_too?[0|1]> :<reason>
SHEDDING <server> OFF
```

Reason is only logged in the snote reporting that shedding has been enabled; it is not otherwise available.  The module is here: https://github.com/oftc/oftc-hybrid/blob/develop/modules/m_shedding.c

dwfreed:
Fixed by #137

PR #137, opened 2021-04-08
Create user shedding module based on oftc-hybrid's m_shedding

examknow:
Per #122

Ported from https://github.com/oftc/oftc-hybrid/blob/c4773b14d1d81f84606d5fe9c7cc6e085b140d00/modules/m_shedding.c


extensions/shedding.c
```diff
@@ -0,0 +1,148 @@
...
+
+static const char shed_desc[] = "Enables/disables user shedding.";
+
+static void mo_shedding(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p, int parc, const char *parv[]);
+void user_shedding_main(void *rate);
+void user_shedding_shed(void *unused);
+
+struct Message shedding_msgtab = {
+	"SHEDDING", 0, 0, 0, 0,
+	{mg_unreg, mg_not_oper, mg_ignore, mg_ignore, mg_ignore, {mo_shedding, 3}}
```
edk0, line 56: You need to set RCLIENT_HANDLER if you want remote SHEDDING to work

edk0 (changes requested):
General things:

* should operstoo ever be used? If not, let's drop it
* we should probably ignore kline-exempt users, as drain does
* we'll need a help file

extensions/shedding.c
```diff
@@ -0,0 +1,148 @@
...
+
+	if (hunt_server(client_p, source_p, ":%s SHEDDING %s %s %s :%s", 1,
+		parc, parv) != HUNTED_ISME)
+		return;
+
+	if (!irccmp(parv[2], "OFF"))
+	{
+		sendto_realops_snomask(SNO_GENERAL, L_ALL, "%s disabled user shedding", get_oper_name(source_p));
+		rb_event_delete(user_shedding_main_ev);
+		rb_event_delete(user_shedding_shed_ev);
```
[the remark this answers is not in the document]
examknow: I don't actually know why, but that change causes shedding to never end even after the stop command is given. Is there any specific reason to add that there?

dwfreed (changes requested):
After analysis of the implementation in oftc-hybrid, I determined that its random variance doesn't actually randomize, and results in some misbehavior.  For solanum, this misbehavior is a memory leak of the first "user shedding shed event" event, and the continued use of that event while subsequent events that should have a different period never actually fire, because they are deleted beforehand.  This will need to be redesigned, with consideration on whether we actually care whether each shed user happens some random period, or if it's fine that they happen exactly every input rate.

help/opers/shedding
```diff
@@ -0,0 +1,9 @@
+SHEDDING <server> <rate> :<reason>
+SHEDDING <server> OFF
+
+Turns user shedding on or off on the specified server. When a server is
+shedding, a client will be disconnected from the server at a time period
+defined by rate, until shedding is turned off. Clients with kline_exempt
+are never effected by this feature.
```
jillest, line 7: typo
```suggestion
are never affected by this feature.
```

extensions/m_shedding.c
```diff
@@ -0,0 +1,160 @@
...
+user_shedding_shed(void *unused)
+{
+	rb_dlink_node *ptr;
+	struct Client *client_p;
+
+	RB_DLINK_FOREACH_PREV(ptr, lclient_list.tail)
+	{
+		client_p = ptr->data;
+
+		if (!MyClient(client_p)) /* It could be servers */
```
jillest, line 150: Perhaps `IsClient` or `IsPerson`, since the client must be local because it's on `lclient_list`.

extensions/m_shedding.c
```diff
@@ -0,0 +1,160 @@
...
+			continue;
+		if (IsExemptKline(client_p))
+			continue;
+		exit_client(client_p, client_p, &me, "Connection closed");
```
jillest, line 154: On some servers, this quit message means the user has been banned from the server.

extensions/m_shedding.c
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if (parc != 4 && !(parc == 3 && irccmp(parv[2], "OFF") == 0))
+	{
+		sendto_one(source_p, form_str(ERR_NEEDMOREPARAMS),
+			me.name, source_p->name, "SHEDDING");
+		return;
+	}
+
+	if (hunt_server(client_p, source_p, ":%s SHEDDING %s %s %s :%s", 1,
```
jillest, line 101: The common way this kind of thing is implemented is using ENCAP. One advantage of this is that intermediate servers need not know about this feature to propagate it correctly. Also, it seems like this format string has too many placeholders for the `OFF` case.

ENCAP would also enable allowing wildcards here easily, but I think that should not be supported since it would be too easy to enable shedding on the whole network.

extensions/m_shedding.c
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * SHEDDING <server> OFF - disable shedding
+ * SHEDDING <server> <approx_seconds_per_userdrop> :<reason>
+ * (parv[#] 1        2                             3)
+ *
+ */
+static void
+mo_shedding(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p, int parc, const char *parv[])
+{
+	if (!HasPrivilege(source_p, "oper:shedding"))
```
jillest, line 88: The common convention is to check this if `source_p` is a local client only. This will happen automatically when using ENCAP and having a separate `me_shedding`.

extensions/m_shedding.c
```diff
@@ -0,0 +1,160 @@
...
+		return;
+	}
+
+	sendto_realops_snomask(SNO_GENERAL, L_ALL | L_NETWIDE, "%s enabled user shedding (interval: %d seconds, reason: %s)",
+		get_oper_name(source_p), rate, parv[3]);
+
+	rate -= (rate/5);
+	rb_event_delete(user_shedding_main_ev);
+	user_shedding_main_ev = NULL;
+	user_shedding_main_ev = rb_event_add("user shedding main event", user_shedding_main, NULL, rate);
```
jillest, line 129: The logic with two events is leaky and feels unnecessary complex. Is randomness really needed here?

extensions/m_shedding.c
examknow: Are you suggesting that we tell users they have been shed from a server?

dwfreed: There's no harm in telling them, imo.  It's going to be pretty obvious regardless.  oftc-hybrid uses "Server closed connection" for shedding, and that is the only use of that particular quit message.

extensions/m_shedding.c
```diff
@@ -0,0 +1,190 @@
...
+
+static void
+me_shedding(struct MsgBuf *msgbuf, struct Client *client_p, struct Client *source_p, int parc, const char *parv[])
+{
+	if(!IsPerson(source_p))
+		return;
+
+	set_shedding_state(source_p, parv[1], parv[2]);
+
+	return;
```
dwfreed, line 169:
```suggestion
```

extensions/m_shedding.c
```diff
@@ -0,0 +1,190 @@
...
+#include "s_conf.h"
+#include "s_serv.h"
+#include "s_newconf.h"
+#include "messages.h"
+#include "numeric.h"
+
+#define SHED_RATE_MIN 5
+
+static int rate = 60;
+
```
dwfreed, line 43:
```suggestion
```

extensions/m_shedding.c
```diff
@@ -0,0 +1,190 @@
...
+moddeinit(void)
+{
+	rb_event_delete(user_shedding_ev);
+}
+
+DECLARE_MODULE_AV2(shed, NULL, moddeinit, shedding_clist, NULL, NULL, NULL, NULL, shed_desc);
+
+static void
+set_shedding_state(struct Client *source_p, const char *chr, const char *reason)
+{
```
dwfreed, line 69:
```suggestion
{
	int rate;
```

dwfreed (changes requested):

extensions/m_shedding.c
```diff
@@ -0,0 +1,190 @@
...
+contains_wildcard(const char *p)
+{
+    while(*p) {
+        if (*p == '*')
+            return true;
+        p++;
+    }
+    return false;
+}
+
```
dwfreed, line 105:
```suggestion
```

dwfreed (approved):

PR #142, opened 2021-04-11
Allow temporary K-lines to extend shorter ones

edk0:
I've also made /kline delete any matching temporary non-propagated K-lines (as effectively happened for propagated ones already) since otherwise this would make it very easy to add duplicates, and made /unkline able to delete multiple matches—while I don't know that there's any way to add unlimited duplicate K-lines, one can add two (a propagated K-line and a longer local K-line, for instance).

The propagated K-line logic already handles extensions adequately; `replace_old_ban` will ensure they win out over the extendee. You just couldn't tell /kline to do them until now.

### DIFF
--- a/modules/m_kline.c
+++ b/modules/m_kline.c
@@ -81,6 +81,7 @@ static void apply_prop_kline(struct Client *source_p, struct ConfItem *aconf,
 static bool already_placed_kline(struct Client *, const char *, const char *, int);
 
 static void handle_remote_unkline(struct Client *source_p, const char *user, const char *host);
+static void remove_superseded_klines(const char *user, const char *host);
 static void remove_permkline_match(struct Client *, struct ConfItem *);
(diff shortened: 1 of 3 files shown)


modules/m_kline.c
```diff
@@ -424,19 +430,22 @@ mo_unkline(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *sour
 		return;
 	}
 
-	if(aconf->lifetime)
+	do
```
glguy, line 433: This sequence kills the server:

```
KLINE 1000 10.0.0.1 :a global ban
UNKLINE 10.0.0.1 ON this.server
```

because there is nothing to end the while loop. That same global kline keeps coming up and never getting removed.

glguy (changes requested):

glguy (approved):

PR #143, opened 2021-04-11
m_whois: put opername/privset in RPL_WHOISOPERATOR

### LINKED ISSUE #100: fold RPL_WHOISSPECIAL for oper/privset in to RPL_WHOISOPERATOR
I imagine we can forego configured oper/admin strings in `RPL_WHOISOPERATOR` when the person doing it has the permission to see the oper/privset info of the target, and as such we can stop using a vaguely contested numeric (`RPL_WHOISSPECIAL`, https://bugs.kde.org/show_bug.cgi?id=130588) and also forego needless information in whois

```
:server 313 jess jess :is an IRC Operator
:server 320 jess jess :is opered as jess, privset sandcat
```
can just become
```
:server 313 jess jess :is opered as jess, privset sandcat
```

edk0:
Show just the opername if you're an oper the target is not hidden from and you don't have oper:privs.

close #100

### DIFF
--- a/modules/m_whois.c
+++ b/modules/m_whois.c
@@ -293,25 +293,32 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
 		sendto_one_numeric(source_p, RPL_AWAY, form_str(RPL_AWAY),
 				   target_p->name, target_p->user->away);
 
-	if((!ConfigFileEntry.hide_opers_in_whois || IsOper(source_p)) && SeesOper(target_p, source_p))
-	{
-		sendto_one_numeric(source_p, RPL_WHOISOPERATOR, form_str(RPL_WHOISOPERATOR),
-				   target_p->name,
-				   IsService(target_p) ? ConfigFileEntry.servicestring :
-				   (IsAdmin(target_p) ? GlobalSetOptions.adminstring :
-				    GlobalSetOptions.operstring));
-	}
-
-	if(!EmptyString(target_p->user->opername) && IsOper(target_p) && (target_p == source_p || HasPrivilege(source_p, "oper:privs")))
+	if (!EmptyString(target_p->user->opername) && IsOper(target_p))
 	{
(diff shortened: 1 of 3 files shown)


modules/m_whois.c
```diff
@@ -293,25 +293,32 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+				privset = target_p->user->privset->name;
+			snprintf(buf, sizeof buf, "is opered as %s, privset %s", target_p->user->opername, privset);
+			s = buf;
+		}
+		else if (IsOper(source_p) && SeesOper(target_p, source_p))
+		{
+			snprintf(buf, sizeof buf, "is opered as %s", target_p->user->opername);
+			s = buf;
+		}
+		else if ((!ConfigFileEntry.hide_opers_in_whois || IsOper(source_p)) && SeesOper(target_p, source_p))
```
dwfreed, line 313: `IsOper()` test here is redundant?  The branch above would always be used instead?

edk0: Oops.

modules/m_whois.c
```diff
@@ -293,25 +293,32 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+				privset = target_p->user->privset->name;
+			snprintf(buf, sizeof buf, "is opered as %s, privset %s", target_p->user->opername, privset);
+			s = buf;
+		}
+		else if (IsOper(source_p) && SeesOper(target_p, source_p))
+		{
+			snprintf(buf, sizeof buf, "is opered as %s", target_p->user->opername);
+			s = buf;
+		}
+		else if (!ConfigFileEntry.hide_opers_in_whois && SeesOper(target_p, source_p))
```
glguy, line 313: I think this case should be outside the test for there being an `opername` so that it will fire even for network services that don't use opername

glguy (approved):

PR #147, opened 2021-04-16
channel: flood warnings for privmsg, not notice

### LINKED ISSUE #146: Flood throttling is not notifying the flooder
I cannot figure out why, but when solanum [throttles a message](https://github.com/solanum-ircd/solanum/blob/a92275551218402bf4969a7fe613cab27416c4ef/modules/core/m_message.c#L884-L903), it does not send the `*** Message to #channel throttled due to flooding` error message but it does trigger a possible flooder snote. I was able to replicate this on two solanum networks with both opered and non-opered clients.

edk0:
fixes #146

### DIFF
--- a/include/channel.h
+++ b/include/channel.h
@@ -32,6 +32,7 @@
 #define MAXMODEPARAMSSERV 10
 
 #include <setup.h>
+#include "hook.h"
 
 struct Client;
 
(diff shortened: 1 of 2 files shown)


examknow (approved):
that did the trick!

glguy (approved):
Change matches old behavior.

The new enum has 4 cases: NOTICE, PRIVMSG, PART, COUNT. Assuming we want flood notices send for the 3 non-NOTICE cases, this change is good. I could also imagine the right test could be: `msgtype == MESSAGE_TYPE_PRIVMSG`.

PR #152, opened 2021-04-26
add separate priv (oper:message) for walking over CALLERID (umode +g)

jesopo:
`oper:general` seems too imprecise for granting this


NEWS.md
```diff
@@ -41,6 +41,7 @@ bolded warnings in the full release notes below.
 - **Breaking:** helpops now uses +h instead of +H
 - **Breaking:** sno\_whois and the spy\_ extensions have been removed
 - **Breaking:** Using /wallops now requires the oper:wallops privilege instead of oper:massnotice
+- **Breaking:** Opers wanting to override CALLERID (usermode +g) need oper:message
```
edk0, line 44: I don't think we need a news entry since it already says all privset configs will be broken

edk0 (commented):
I don't think we need a NEWS entry given it already says all privs will need to be rewritten

NEWS.md
edk0: ... I thought I cancelled the comment and then added this review. Oops.

edk0 (approved):

modules/um_callerid.c
```diff
@@ -103,7 +103,7 @@ allow_message(struct Client *source_p, struct Client *target_p)
 		return true;
 
 	/* XXX: controversial?  allow opers to send through +g */
```
aaronmdjones, line 105: Should this comment still be here now?

PR #160, opened 2021-05-19
Make services always allowed to bypass umode +g

examknow:


aaronmdjones:
This was already fixed in https://github.com/edk0/solanum/commit/b8b30d2f0f712730fe2e81a686e84806223562d5 and the fix deployed to the network. The commits will be tidied up and put into the repository proper when we're not busy with network administration.

PR #169, opened 2021-05-25
Use plain language in +r/+R error messages

mquin:


mniip (commented):
What about `NUMERIC_STR_486`?
And also maybe we should mention NickServ explicitly?

edk0 (approved):

PR #178, opened 2021-05-31
Port m_invex_regonly from ircd-seven

mquin:
This module allows +I to be used to bypass +r (registered only) as well as +i (invite only).

### DIFF
--- a/extensions/invex_regonly.c
+++ b/extensions/invex_regonly.c
@@ -26,12 +26,12 @@ h_can_join(hook_data_channel *data)
 	struct matchset ms;
 	rb_dlink_node *ptr;
 	
-	matchset_for_client(source_p, &ms);
-
 	if(data->approved != ERR_NEEDREGGEDNICK)
 		return;
 	if(!ConfigChannel.use_invex)
 		return;
+
+	matchset_for_client(source_p, &ms);
 
 	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
 	{
@@ -43,4 +43,3 @@ h_can_join(hook_data_channel *data)
 	if(ptr != NULL)
 		data->approved=0;
 }
-
--- a/extensions/invex_regonly.c
+++ b/extensions/invex_regonly.c
@@ -35,10 +35,10 @@ h_can_join(hook_data_channel *data)
 
 	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
 	{
-	invex = ptr->data;
-	if (matches_mask(&ms, invex->banstr) ||
-			match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
-				break;
+		invex = ptr->data;
+		if (matches_mask(&ms, invex->banstr) ||
+				match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
+			break;
 	}
 	if(ptr != NULL)
 		data->approved=0;
--- a/extensions/invex_regonly.c
+++ b/extensions/invex_regonly.c
@@ -38,8 +38,9 @@ h_can_join(hook_data_channel *data)
 		invex = ptr->data;
 		if (matches_mask(&ms, invex->banstr) ||
 				match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
+		{
+			data->approved = 0;
 			break;
+		}
 	}
-	if(ptr != NULL)
-		data->approved=0;
 }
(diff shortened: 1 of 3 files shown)


extensions/m_invex_regonly.c
```diff
@@ -0,0 +1,47 @@
...
+{
+	struct Client *source_p = data->client;
+	struct Channel *chptr = data->chptr;
+	struct Ban *invex = NULL;
+	struct matchset ms;
+	rb_dlink_node *ptr;
+	
+	matchset_for_client(source_p, &ms);
+
+	if(data->approved == ERR_NEEDREGGEDNICK) {
```
edk0, line 31: The indentation has gone a bit funny here. Also it might look better if we return if `data->approved != ERR_NEEDREGGEDNICK`, then the rest of the body doesn't need to be indented anyway.

extensions/m_invex_regonly.c
```diff
@@ -0,0 +1,47 @@
+/*
+ * m_invex_regonly.c Allow invite exemptions to bypass registered-only (+r)
```
edk0, line 2: I guess this could just be called `invex_regonly`.

edk0 (commented):
Thanks 🍰

edk0 (approved):

extensions/invex_regonly.c
```diff
@@ -0,0 +1,46 @@
...
+	struct Ban *invex = NULL;
+	struct matchset ms;
+	rb_dlink_node *ptr;
+	
+	matchset_for_client(source_p, &ms);
+
+	if(data->approved != ERR_NEEDREGGEDNICK)
+		return;
+	if(!ConfigChannel.use_invex)
+		return;
```
dwfreed, line 34: Put the matchset creation after the not-for-us tests; it's not expensive, but it's not free either.

```suggestion
	if(data->approved != ERR_NEEDREGGEDNICK)
		return;
	if(!ConfigChannel.use_invex)
		return;

	matchset_for_client(source_p, &ms);
```

extensions/invex_regonly.c
```diff
@@ -0,0 +1,46 @@
...
+
+	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
+	{
+	invex = ptr->data;
+	if (matches_mask(&ms, invex->banstr) ||
+			match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
+				break;
+	}
+	if(ptr != NULL)
+		data->approved=0;
```
dwfreed: Move setting `data->approved` inside the if in the loop, just so we won't have to test again.  The second test might get optimized away anyway, but it's better to not have to rely on it.  Also includes some indentation fixes.

```suggestion
	{
		invex = ptr->data;
		if (matches_mask(&ms, invex->banstr) ||
			match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
		{
			data->approved = 0;
			break;
		}
	}
```

extensions/invex_regonly.c
```diff
@@ -0,0 +1,46 @@
...
+
+	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
+	{
+		invex = ptr->data;
+		if (matches_mask(&ms, invex->banstr) ||
+				match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
+			break;
+	}
+	if(ptr != NULL)
+		data->approved=0;
```
dwfreed, line 44: Move setting `data->approved` within the if, so we don't have to test again.  The second test might get optimized away, but it's better to not have to rely on that.

```suggestion
		if (matches_mask(&ms, invex->banstr) ||
				match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
		{
			data->approved = 0;
			break;
		}
	}
```

dwfreed (changes requested):
Some minor things

jillest:
Consideration: just make this the standard behaviour? I don't feel very strongly that the existing interaction between +r and +I is the right solution.

ncoevoet:
+e yes, but +I ?

dwfreed (approved):

mquin:
> +e yes, but +I ?

+r isn't a ban, so it doens't make sense for +e to be an exemption for it

/invite bypasses +r, but not +b - this extends that behaviour to +I

The expected use case for this is allowing unidentified users from particular gateways, and you'd likely still want to be able to ban individual users selectively.

dwfreed:
I approved, but I also have no objection to making this the normal behavior.

mquin nailed the logic behind making this +I instead of +e

PR #204, opened 2021-06-12
msgbuf: use only relevant caps for the cache key

edk0:
Given cache->overall_capmask was already computed, but not used, I haved a feeling that this was always planned but the original implementers forgot about it. Without this change, messages to large channels with diverse clients consistently cause _many_ cache evictions; with it they never do.

Reviewers: the `caps` parameter we're modifying eventually makes its way to `msgbuf_unparse_tags`, where it's &ed with `msgbuf->tags[foo].capmask`, the same field whose | makes up overall_capmask. Given that, I think any bits in caps that aren't in overall_capmask will never be used anyway.


glguy (approved):
Good spot

PR #205, opened 2021-06-13
m_stats: when viewing self using /STATS l|L, show real hostname or IP

### LINKED ISSUE #191: "/stats l" and "/stats L" (without <nick>) both return IP 255.255.255.255 when cloaked
for  "vHost: user/joerg" **cloaked users**, the
command 
`/stats L`  and
`/stats l`
both return IP 255.255.255.255
`[211] joerg joerg[~saturn@255.255.255.255] 99 445 51 8 24 30767679 -`

EXPECTED:
probably real IP for `/stats L`  and vHost cloak for `/stats l`

megubyte:
Closes #191.

It probably makes sense to show you your own IP and/or hostname when using /STATS l or /STATS L. You can't see other people anyway, and this brings the behaviour in line with /WHOIS (namely the line that shows you your real IP).

### DIFF
--- a/modules/m_stats.c
+++ b/modules/m_stats.c
@@ -1654,7 +1654,7 @@ stats_l_client(struct Client *source_p, struct Client *target_p,
 	else
 	{
 		sendto_one_numeric(source_p, RPL_STATSLINKINFO, Lformat,
-				   show_ip(source_p, target_p) ?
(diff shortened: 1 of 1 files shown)


edk0:
I'm not certain whether #191 is really a bug, but setting that aside, I don't think we should reveal iline-spoofed IPs here.

PR #216, opened 2021-06-17
Remove unneeded ugly hack for 32-bit Solaris

jailbird777:
- The official Sun/Oracle solution is to use the extendedFILE(5)
  mechanism, which works around the limitation.
  https://docs.oracle.com/cd/E18752_01/html/816-5175/extendedfile-5.html
- Add a quick HOWTO to the README.md


kaniini:
Looks fine... but do we really care about 32-bit solaris?

kaniini (approved):

jailbird777:
I personally don’t care about 32-bit. I just ran into the hack while debugging the listener issue since it unnecessarily got used in 64-bit builds also.

edk0 (approved):

Issue #214, opened 2021-06-17
add solanum.chat/oper capability, message tag


edk0:
should add a `@solanum.chat/opername=foo` tag to messages from an oper whose oper block is named `foo`. The `cap_realhost` extension can serve as an example. As with realhost, there will need to be a hidden version of the capability that is automatically set for opers who can see through oper hiding.

edk0:
revision after thinking about it for a few more seconds: let's make it `@solanum.chat/oper` and include a name if we know the name.

we'll probably need two hidden caps -- one for opers who can see through oper hiding (i.e. they have `auspex:oper`), and one for all other opers. other users will never see opernames, but they would see the oper status of non-hidden opers, so we *could* add yet another hidden cap for them... but that's getting quite expensive in terms of capability bits.

SadieCat:
FWIW if you'd be interested in taking this to IRCv3 I'd also be interested in implementing it.

edk0:
> FWIW if you'd be interested in taking this to IRCv3 I'd also be interested in implementing it.

yeah, I don't see why not

examknow:
We'd also need to check if the oper has `oper:hidden` when sending the tag to people who don't have `oper:auspex` even if oper hiding isn't configured server wide.

kaniini:
We will be glad to carry this in Ophion as well.

PR #217, opened 2021-06-19
Add `solanum.chat/oper` capablity

examknow:
Per #214. I understand the IRCv3 people might want to get in on this too so we may need to end up changing this depending on how that goes.

### DIFF
--- a/extensions/cap_oper.c
+++ b/extensions/cap_oper.c
@@ -1,0 +1,152 @@
+/*
+ *  Copyright (C) 2021 David Schultz <me@zpld.me>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


extensions/cap_oper.c
```diff
@@ -0,0 +1,152 @@
...
+	client->localClient->caps &= ~CLICAP_OPER_AUSPEX;
+	client->localClient->caps &= ~CLICAP_OPER_JUSTOPER;
+	client->localClient->caps &= ~CLICAP_OPER_NORMAL;
+
+	if (client->localClient->caps & CLICAP_OPER && HasPrivilege(client, "auspex:oper"))
+	{
+		/* if the client is an oper with auspex, let them see everything */
+		client->localClient->caps |= CLICAP_OPER_AUSPEX;
+	}
+	else if (ConfigFileEntry.hide_opers)
```
edk0, line 100: I think this needs to be in `cap_oper_outbound_msgbuf`, since it's a configuration setting that could change at runtime.

edk0 (changes requested):
I have one minor change, but this looks very good. Thanks 🍰

edk0 (dismissed):

edk0 (approved):

Issue #151, opened 2021-04-24
add a way to disable idle time on your connection


jesopo:
some way to hide
- idle time in `WHOIS` output
- idle time in `TRACE` output

maybe a usermode?

PR #220, opened 2021-06-24
Add umode +I to allow users to hide their idle time

examknow:
Per #151 - Fires a hook whenever someone runs `TRACE` or `WHOIS` that is used by `extensions/hide_idle_time` to determine whether or not the user is allowed to see idle times. Users whois/tracing themselves or opers with the `auspex:usertimes` privilege are always allowed to see idle times, regardless of whether or not umode +I is set.


aaronmdjones:
I'm thinking it would be nice for this to also add a privilege, say, `auspex:usertimes`, that lets opers see a user's connect/idle timestamps without having to operspy.

(If you do, don't forget to document it in reference.conf, under the section "added by extensions")

examknow:
> 
> 
> I'm thinking it would be nice for this to also add a privilege, say, `auspex:usertimes`, that lets opers see a user's connect/idle timestamps without having to operspy.

I think that's a good idea

examknow:
> 
> 
> I'm thinking it would be nice for this to also add a privilege, say, `auspex:usertimes`, that lets opers see a user's connect/idle timestamps without having to operspy.

I think that's a good idea

extensions/hide_idle_time.c
```diff
@@ -0,0 +1,69 @@
...
+
+static void
+h_huc_doing_trace_or_whois(void *data_)
+{
+	hook_data_client_approval *data = data_;
+
+	if (data->target->umodes & user_modes['I'])
+	{
+		if ((data->client == data->target) || HasPrivilege(data->client, "auspex:usertimes"))
+			data->approved = 1;
```
edk0, line 48: We probably shouldn't approve here if it's already been unapproved by something else.

extensions/hide_idle_time.c
```diff
@@ -0,0 +1,69 @@
...
+#include "s_user.h"
+#include "s_newconf.h"
+
+static const char hide_desc[] = "Provides user mode +I to hide a user's idle time";
+
+static void h_huc_doing_trace_or_whois(void *);
+
+mapi_hfn_list_av1 huc_hfnlist[] = {
+	{ "doing_whois_show_idle", (hookfn) h_huc_doing_trace_or_whois },
+	{ "doing_trace_show_idle", (hookfn) h_huc_doing_trace_or_whois },
```
edk0, line 36:
```suggestion
	{ "doing_whois_show_idle", h_huc_doing_trace_or_whois },
	{ "doing_trace_show_idle", h_huc_doing_trace_or_whois },
```

modules/m_whois.c
```diff
@@ -373,10 +375,23 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+		call_hook(doing_whois_show_idle_hook, &hdata_showidle);
+
+		if (hdata_showidle.approved || operspy)
+		{
+			/* operspies are always allowed to see idle time */
+			sendto_one_numeric(source_p, RPL_WHOISIDLE, form_str(RPL_WHOISIDLE),
+					   target_p->name,
+					   (long)(rb_current_time() - target_p->localClient->last),
+					   (unsigned long)target_p->localClient->firsttime);
+		}
```
edk0, line 394: Would it make sense to show the connect time even if the idle time is hidden? I don't think it gives much away.

edk0 (dismissed):

modules/m_whois.c
examknow: If we did show the connect time, we'd have to send some garbage for the idle time which could be confusing. I just decided to hide the whole thing since that's how unrealircd behaves.

jesopo:
I think it's a bad idea to hide connected time. either send idle time as time since they connected, or `0`

examknow:
> 
> 
> I think it's a bad idea to hide connected time. either send idle time as time since they connected, or `0`

In that case I think sending `0` is most appropriate.

jesopo:
sgtm

jillest:
STATS l/L also show the connection idle time, and send/recv messages/bytes gives related information about the connection.

modules/m_whois.c
```diff
@@ -373,10 +375,20 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+
+		call_hook(doing_whois_show_idle_hook, &hdata_showidle);
+
+		/* operspies are always allowed to see idle time */
 		sendto_one_numeric(source_p, RPL_WHOISIDLE, form_str(RPL_WHOISIDLE),
-				   target_p->name,
-				   (long)(rb_current_time() - target_p->localClient->last),
-				   (unsigned long)target_p->localClient->firsttime);
+			   target_p->name,
+			   hdata_showidle.approved || operspy ? (long)(rb_current_time() - target_p->localClient->lasttime) : 0,
```
dwfreed, line 390: I disagree with including operspy here; there's already a separate priv for it, if a network wants operspy folk to see it too, they can give that priv as well.

Also, this needs to stay using last, not lasttime.  Last is the last time user sent a PRIVMSG, lasttime is the last time the user sent any line, and is only used for ping calculations.

```suggestion
			   hdata_showidle.approved ? (long)(rb_current_time() - target_p->localClient->last) : 0,
```

This suggestion is just the last revert, in case Libera disagrees with me about not including operspy here

```suggestion
			   hdata_showidle.approved || operspy ? (long)(rb_current_time() - target_p->localClient->last) : 0,
```

modules/m_whois.c
```diff
@@ -373,10 +375,20 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+		/* fire the doing_whois_show_idle hook to allow modules to tell us whether to show the idle time */
+		hook_data_client_approval hdata_showidle;
+
+		hdata_showidle.client = source_p;
+		hdata_showidle.target = target_p;
+		hdata_showidle.approved = 1;
+
+		call_hook(doing_whois_show_idle_hook, &hdata_showidle);
+
+		/* operspies are always allowed to see idle time */
```
dwfreed, line 387: Removing this comment to go along with my comment about not lumping operspy in with this.

```suggestion
```

dwfreed (changes requested):
There is also an idle time field in WHOX that this does not cover at present.

modules/m_whois.c
examknow: After discussion on IRC, we've decided to drop the operspy bit here.

extensions/hide_idle_time.c
```diff
@@ -0,0 +1,71 @@
...
+	{ "doing_who_show_idle", h_huc_doing_idle_time_hook },
+	{ NULL, NULL }
+};
+
+static void
+h_huc_doing_idle_time_hook(void *data_)
+{
+	hook_data_client_approval *data = data_;
+
+	if (data->target->umodes & user_modes['I'])
```
dwfreed, line 47: This is missing the shortcut out if approved is already 0:

```suggestion
	if (data->approved == 0)
		return;

	if (data->target->umodes & user_modes['I'])
```

modules/m_whois.c
```diff
@@ -373,10 +375,23 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
-				   target_p->name,
-				   (long)(rb_current_time() - target_p->localClient->last),
-				   (unsigned long)target_p->localClient->firsttime);
+			   target_p->name,
+			   hdata_showidle.approved ? (long)(rb_current_time() - target_p->localClient->last) : 0,
+			   (unsigned long)target_p->localClient->firsttime);
+
+		if (hdata_showidle.approved == 2 || hdata_showidle.approved == 0)
+			/* if the target has hidden their idle time, notify the source */
+			sendto_one_numeric(source_p, RPL_WHOISTEXT, "%s :is using a private idle time", target_p->name);
```
dwfreed, line 394: This should use the form_str:

```suggestion
			sendto_one_numeric(source_p, RPL_WHOISTEXT, form_str(RPL_WHOISTEXT), target_p->name, "is using a private idle time");
```

dwfreed (changes requested):

dwfreed (dismissed):

extensions/Makefile.am
```diff
@@ -69,6 +69,7 @@ extension_LTLIBRARIES =		\
   identify_msg.la		\
   cap_realhost.la		\
   invex_regonly.la		\
+  hide_idle_time.la             \
```
glguy, line 72:
```suggestion
  hide_idle_time.la		\
```

extensions/hide_idle_time.c
```diff
@@ -0,0 +1,74 @@
...
+			data->approved = 2;
+	}
+}
+
+static int
+_modinit(void)
+{
+	user_modes['I'] = find_umode_slot();
+	construct_umodebuf();
+	return 0;
```
glguy, line 64: Other umode modules (like um_regonlymsg.c) handle failure of find_umode_slot(). Perhaps this should, too?

modules/m_whois.c
```diff
@@ -373,10 +375,23 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
...
+
 		sendto_one_numeric(source_p, RPL_WHOISIDLE, form_str(RPL_WHOISIDLE),
-				   target_p->name,
-				   (long)(rb_current_time() - target_p->localClient->last),
-				   (unsigned long)target_p->localClient->firsttime);
+			   target_p->name,
+			   hdata_showidle.approved ? (long)(rb_current_time() - target_p->localClient->last) : 0,
+			   (unsigned long)target_p->localClient->firsttime);
+
+		if (hdata_showidle.approved == 2 || hdata_showidle.approved == 0)
```
glguy, line 392: solanum has been doing better about using enums rather than multiple magic numbers. This seems like a nice place to use something like:

```c
enum whois_idle_approval
{
	WHOIS_IDLE_HIDE = 0,
	WHOIS_IDLE_SHOW = 1,
	WHOIS_IDLE_AUSPEX = 2
};
```

extensions/hide_idle_time.c
```diff
@@ -0,0 +1,74 @@
+/*
```
glguy, line 1: solanum isn't completely consistent here, but this should be `um_hide_idle_time.c` to follow the scheme started in `modules/` or `umode_` to follow `umode_ctcp.c`. We've got a good thing kind of going with `sno_`, `extb_`, `m_` for command things.

I'm happy to lose this one, but I feel like we should be working toward more order instead of less.

modules/m_stats.c
```diff
@@ -1619,19 +1621,27 @@ stats_l_client(struct Client *source_p, struct Client *target_p,
...
 				     get_client_name(target_p, SHOW_IP) :
 				     get_client_name(target_p, HIDE_IP)) :
 				    get_client_name(target_p, MASK_IP),
-				    (int) rb_linebuf_len(&target_p->localClient->buf_sendq),
-				    (int) target_p->localClient->sendM,
-				    (int) target_p->localClient->sendK,
-				    (int) target_p->localClient->receiveM,
-				    (int) target_p->localClient->receiveK,
+				    hdata_showidle.approved ? (int) rb_linebuf_len(&target_p->localClient->buf_sendq): 0,
+				    hdata_showidle.approved ? (int) target_p->localClient->sendM : 0,
```
glguy, line 1639: `0` is a valid value. I'd prefer to see a distinct value like `*` or `-1` here when it was hidden from view. This command is already restricted in who can run it. We shouldn't make the output *wrong* for operators without the privilege.

examknow: Hiding `317` entirely was the original behavior but then was changed per https://github.com/solanum-ircd/solanum/pull/220#discussion_r659179325 and https://github.com/solanum-ircd/solanum/pull/220#issuecomment-879386882. We eventually ended up putting the `is using a private idle time` bit into the `WHOIS` response to make that less confusing. As far as `stats` is concerned, I think it does make sense to use `-1` here.

modules/m_whois.c
```diff
@@ -60,15 +60,24 @@ struct Message whois_msgtab = {
...
 mapi_clist_av1 whois_clist[] = { &whois_msgtab, NULL };
 mapi_hlist_av1 whois_hlist[] = {
 	{ "doing_whois",			&doing_whois_hook },
 	{ "doing_whois_global",			&doing_whois_global_hook },
 	{ "doing_whois_channel_visibility",	&doing_whois_channel_visibility_hook },
+	{ "doing_whois_show_idle",		&doing_whois_show_idle_hook },
 	{ NULL, NULL }
 };
 
+enum whois_idle_approval
```
glguy, line 74: This is useful, but it only gets used in this file while the magic numbers get used in all the other files?

modules/m_trace.c
```diff
@@ -381,13 +383,22 @@ report_this_status(struct Client *source_p, struct Client *target_p)
...
+			call_hook(doing_trace_show_idle_hook, &hdata_showidle);
+
 			sendto_one_numeric(source_p,
 					SeesOper(target_p, source_p) ? RPL_TRACEOPERATOR : RPL_TRACEUSER,
 					SeesOper(target_p, source_p) ? form_str(RPL_TRACEOPERATOR) : form_str(RPL_TRACEUSER),
 					class_name, name,
 					show_ip(source_p, target_p) ? ip : empty_sockhost,
-					(unsigned long)(rb_current_time() - target_p->localClient->lasttime),
-					(unsigned long)(rb_current_time() - target_p->localClient->last));
+					hdata_showidle.approved ? (unsigned long)(rb_current_time() - target_p->localClient->lasttime) : 0,
```
glguy, line 400: `stats l` uses -1 to show it's hiding the value, but this uses 0 still

glguy:
I've built in run this myself and can confirm the functionality is working.

extensions/umode_hide_idle_time.c
```diff
@@ -0,0 +1,80 @@
...
+h_huc_doing_idle_time_hook(void *data_)
+{
+	hook_data_client_approval *data = data_;
+
+	if (data->approved == 0)
+		return;
+
+	if (data->target->umodes & user_modes['I'])
+	{
+		if ((data->client != data->target) && !HasPrivilege(data->client, "auspex:usertimes"))
```
glguy, line 53: always do the priv check once instead of either once or twice.

```c
if (HasPrivilege(data->client, "auspex:usertimes"))
			data->approved = 2;
else if (data->client != data->target)
			data->approved = 0;
```

modules/m_whois.c
examknow: Did you have an idea about which header file would be best? I was thinking `hook.h`.

modules/m_trace.c
examknow: IRC discussion in a nutshell:
```
11:43:30 < launchd> does anybody have opinions on https://github.com/solanum-ircd/solanum/pull/220#discussion_r720620067
11:59:02 < dwfreed> hexley: I vote 0
11:59:33 < dwfreed> whatever's reading this may not be expecting a negative number
11:59:53 < dwfreed> and moreso may not be expecting a non-number
12:49:10 < spb> if there's another part of the protocol that already uses 0 for 'idle time not known', then i'm not sure how much of a debate there is to be had about what to use in the new case
```

jesopo:
on second/third/fouth thought, it might be better to make a hidden idletime be the amount of time you've been connected for, instead of `0`, otherwise someone might see the `0` and assume you're active right now, without realising it's a fake number

examknow:
> on second/third/fouth thought, it might be better to make the hidden idletime be the amount of time you've been connected for, instead of `0`, otherwise someone might see the `0` and assume you're active right now, without realising it's a fake number

I think there will always be a potential to confuse people, no matter how we do this. I can see someone being confused and thinking that someone is MIA because the idle time is so huge. Most clients will show the private idle time notice, but there's not much we can do about the ones that don't.

dwfreed (approved):

PR #225, opened 2021-06-27
reference.conf: Make it clearer connect {} blocks are reciprocal

ProgVal:


doc/reference.conf
```diff
@@ -546,12 +546,17 @@ operator "god" {
...
+ * outgoing. */
 connect "irc.uplink.com" {
 	/* the name must go above */
 
-	/* host: the host or IP to connect to.  If a hostname is used it
-	 * must match the reverse dns of the server.
+	/* host: the host or IP to connect to.
+     *
+     * It is also used to validate incoming connections. If a hostname
+     * is used, it must match the reverse dns of the server.
```
dwfreed, line 559: This has mixed indentation

```suggestion
	/* host: the host or IP to connect to.
	 *
	 * It is also used to validate incoming connections. If a hostname
	 * is used, it must match the reverse dns of the server.
```

dwfreed (changes requested):

doc/reference.conf
```diff
@@ -546,12 +546,17 @@ operator "god" {
 	privset = "admin";
 };
 
-/* connect {}: controls servers we connect to (OLD C:, N:, H:, L:) */
+/* connect {}: controls servers we connect with (OLD C:, N:, H:, L:).
+ *
+ * This configuration is used whether connections are incoming or
+ * outgoing. */
```
dwfreed, line 552: Missed that this end-comment should be on a line by itself.  I'll make this change myself.

```suggestion
 * outgoing.
 */
```

doc/reference.conf
```diff
@@ -546,14 +546,20 @@ operator "god" {
...
-/* connect {}: controls servers we connect to (OLD C:, N:, H:, L:) */
+/* connect {}: controls servers we connect with (OLD C:, N:, H:, L:).
+ *
+ * This configuration is used whether connections are incoming or
+ * outgoing.
+ */
 connect "irc.uplink.com" {
 	/* the name of the other server must go above. It should match the
 	 * other server's name in its serverinfo {} block, and does not
 	 * need to be an actual hostname. */
```
dwfreed, line 557: Gonna also fix this since I noticed it, because gross.

```suggestion
	 * need to be an actual hostname.
	 */
```

dwfreed (approved):

PR #233, opened 2021-07-09
Remove Windows support

jailbird777:


librb/include/rb_dictionary.h
```diff
@@ -157,24 +157,12 @@ extern unsigned int rb_dictionary_size(rb_dictionary *dtree);
 void rb_dictionary_stats(rb_dictionary *dict, void (*cb)(const char *line, void *privdata), void *privdata);
 void rb_dictionary_stats_walk(void (*cb)(const char *line, void *privdata), void *privdata);
 
-#ifndef _WIN32
-
 #define RB_POINTER_TO_INT(x)		((int32_t) (long) (x))
 #define RB_INT_TO_POINTER(x)		((void *) (long) (int32_t) (x))
 
 #define RB_POINTER_TO_UINT(x)		((uint32_t) (unsigned long) (x))
 #define RB_UINT_TO_POINTER(x)		((void *) (unsigned long) (uint32_t) (x))
```
jailbird777, line 164: These could be removed now and basically inlined with the code, but that might make the code less readable in the long run.

heroku-miraheze (dismissed):

aaronmdjones (dismissed):

configure.ac
```diff
@@ -653,15 +635,8 @@ AC_SUBST(SEDOBJ)
 
 if test "$prefix" = "NONE"; then
 	AC_DEFINE_UNQUOTED(IRCD_PREFIX, "$ac_default_prefix", [Prefix where the ircd is installed.])
-
 else
-
-dnl Don't get bitten by Cygwin's stupidity if the user specified
-dnl a custom prefix with a trailing slash
-
-	prefix=`echo $prefix | sed 's/\/$//'`
```
dwfreed, line 662: Let's keep this line, in case someone is crazy enough to use `/` as prefix.  POSIX allows for leading `//` to be treated specially, and I have personally worked in an environment where that is the case (not that I'd run an ircd in that environment, but people do crazy things).

jailbird777: Wow, you're right, I've missed that part of POSIX.   I guess it's for network shares or something? I just readded that line with a better comment.  Thanks!

aaronmdjones (approved):

PR #242, opened 2021-07-11
add SNO_BANNED, snote for it on client k/x-line rejection

### LINKED ISSUE #234: add snomask char to show all new connection d/k/x-line rejections
might be handy to be able to opt in to seeing things bounce off klines/dlines

jesopo:
closes #234


jesopo:
maybe there's good enough reason to extend this to project rejected for not using SASL on a SASL only source

dwfreed:
I would say that could be done in the unauthorized snomask, and should cover any failure to meet their matching auth block requirements, not just SASL (secure comes to mind).

dwfreed (changes requested):
This needs a line in the snomask help file

dwfreed (approved):

Issue #240, opened 2021-07-11
drain_reason config option missing from documentation


meldra:
```
16:59 <launchd> oh for starters, you should add it to the example config file and reference.conf (located in doc/)
17:00 <el> i copied drain_reason so that also needs mentioning there then
17:01 <el> perhaps figure out what else it's missing and then make an issue against that so it doesn't get forgotten
17:02 <launchd> yeah probably, although that's not a default module so idk
17:02 <el> ah
17:02 <launchd> edk: thoughts?
17:08 <edk> adding it to reference.conf would make sense in both cases
```

PR #243, opened 2021-07-11
Add drain_reason to reference.conf

examknow:
Per #240


aaronmdjones:
Because this refers to functionality provided by an optional extension, I think you should clarify what needs to be loaded for it to have any effect.

aaronmdjones (approved):

PR #249, opened 2021-07-16
m_services: delete nick delay on RSFNC

### LINKED ISSUE #248: "Nick/channel is temporarily unavailable" (437) message should not be sent if you're already using that nickname
Scenario that just happened on libera:
* Got disconnected from the network due to a netsplit
* Reconnected and did `/ns regain` before ZNC's keepnick module got to it
* ZNC's keepnick module keeps trying to get the nick (arguably it is buggy too because it should see that I'm already using it)
* My server buffer is spammed with `-- nyuszika7h: Nick/channel is temporarily unavailable`, even though my current nick is already nyuszka7h

dwfreed:
Fixes #248

### DIFF
--- a/modules/m_services.c
+++ b/modules/m_services.c
@@ -248,7 +248,8 @@ doit:
 	del_from_client_hash(target_p->name, target_p);
 
 	/* invalidate nick delay because we're forcing this nick to be used */
-	if((nd = rb_dictionary_retrieve(nd_dict, parv[2])))
+	nd = rb_dictionary_retrieve(nd_dict, parv[2]);
+	if (nd != NULL)
 		free_nd_entry(nd);
 
 	rb_strlcpy(target_p->name, parv[2], NICKLEN);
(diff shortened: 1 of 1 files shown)


modules/m_services.c
```diff
@@ -245,6 +246,11 @@ me_rsfnc(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 			use_id(target_p), parv[2], (long) target_p->tsinfo);
 
 	del_from_client_hash(target_p->name, target_p);
+
+	/* invalidate nick delay because we're forcing this nick to be used */
+	if((nd = rb_dictionary_retrieve(nd_dict, parv[2])))
```
edk0, line 251:
```suggestion
	nd = rb_dictionary_retrieve(nd_dict, parv[2]);
	if (nd != NULL)
```

No need to use the value of an assignment here, IMO.  
<sub><sub>(and no space after a keyword is icky)</sub></sub>

edk0 (approved):

PR #257, opened 2021-08-07
A bunch of fixes

jailbird777:
- NetBSD needs netinet/tcp.h for TCP_NODELAY
- Remove some trailing ;s that cause a few warnings
- getexecname returns const
- Clean up a bunch of cppcheck warnings. I have a feeling that the va_start
  warnings are benign, but might as well clean them up
- There's a bunch of cppcheck warnings about being able to reduce the scope
  of variables, however I didn't include those in this.

### DIFF
--- a/modules/m_whowas.c
+++ b/modules/m_whowas.c
@@ -116,7 +116,7 @@ m_whowas(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 	RB_DLINK_FOREACH(ptr, whowas_list->head)
 	{
 		struct Whowas *temp = ptr->data;
-		if(cur > 0 && rb_linebuf_len(&client_p->localClient->buf_sendq) > sendq_limit)
+		if(rb_linebuf_len(&client_p->localClient->buf_sendq) > sendq_limit)
 		{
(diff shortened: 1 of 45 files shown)


modules/m_whowas.c
```diff
@@ -116,7 +116,7 @@ m_whowas(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 	RB_DLINK_FOREACH(ptr, whowas_list->head)
 	{
 		struct Whowas *temp = ptr->data;
-		if(cur > 0 && rb_linebuf_len(&client_p->localClient->buf_sendq) > sendq_limit)
+		if(rb_linebuf_len(&client_p->localClient->buf_sendq) > sendq_limit)
```
dwfreed, line 119: This removal is wrong.  We still want to send out 1 whowas match even if it exceeds the safe sendq limit.

dwfreed (changes requested):
Please split this into several commits covering specific issues with rationale for each.  It's very hard to review changes when there is no rationale provided.  It's also very hard to review changes to 45 different files at a time, but it's a lot easier when there's distinct commits covering each set of related changes.

PR #258, opened 2021-08-10
MODE +k: Return ERR_INVALIDMODEPARAM instead of stripping invalid chars or truncating

ProgVal:
This is more user-friendly (prevents setting a different key than
what the other requested) and makes the behavior consistent with
InspIRCd and Ergo (and hopefully others soon).

https://github.com/ircdocs/modern-irc/pull/111

### DIFF
--- a/include/messages.h
+++ b/include/messages.h
@@ -203,6 +203,7 @@
 #define NUMERIC_STR_670      ":STARTTLS successful, proceed with TLS handshake"
 #define NUMERIC_STR_671      "%s :%s"
 #define NUMERIC_STR_691      ":%s"
+#define NUMERIC_STR_696      ":%s 696 %s %s %c %s :%s"
 #define NUMERIC_STR_702      ":%s 702 %s %s 0x%lx %s%s :[Version %s] [Description: %s]"
 #define NUMERIC_STR_703      ":%s 703 %s :End of /MODLIST."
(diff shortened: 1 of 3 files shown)


edk0 (changes requested):
Per IRC, I think we should either accept or reject the whole mode command, not error while still actioning the rest of it.

ProgVal:
> Per IRC

Where/when? I think I missed this discussion.


`ERR_INVALIDMODEPARAM` mention which char is causing the issue so client can figure it out.

Both InspIRCd and Ergo do it like this:

```
C -> S: MODE #chan +vk bar :passphrase with spaces
S -> C: :My.Little.Server 696 bar #chan k * :Invalid key mode parameter. Syntax: <key>.
S -> C: :bar!username@127.0.0.1 MODE #chan +v :bar
```

dwfreed:
```
2021-08-10 17:03:24 < edk> yes but it's not that simple
2021-08-10 17:03:41 < edk> currently by the time we know it's an invalid key we've already started actioning the command
2021-08-10 17:03:57 < edk> and i think we should either accept the whole command or reject it
```

Timestamp is UTC; this was said about 10 minutes after you brought up this issue.

ProgVal:
Okay nevermind then I'd rather keep solanum constant with Hybrid than introduce yet another behavior

PR #260, opened 2021-08-12
add SNO_FARCONNECT to the help text

xfnw:
snomask +F seems to be missing from the help text

### DIFF
--- a/help/opers/snomask
+++ b/help/opers/snomask
@@ -1,6 +1,8 @@
 MODE <nick> +s <+|-><modes>
 
-Server notice masks:
+Server notice masks: (? designates that the snomask is provided
+                      by an extension and may not be present
+                      on this server)
 
     SNOMASK     DESCRIPTION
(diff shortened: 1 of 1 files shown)


examknow:
We should probably also mention that this requires `extensions/sno_farconnect` to be loaded.

xfnw:
hmm, the other entries dont mention required extensions

aaronmdjones:
Follow the convention in the `UMODE` help, and put a `?` next to snomasks that require extensions?

help/opers/snomask
```diff
@@ -1,6 +1,8 @@
 MODE <nick> +s <+|-><modes>
 
-Server notice masks:
+Server notice masks: (? designates that the umode is provided
```
aaronmdjones, line 3: snomask, not umode.

aaronmdjones (changes requested):

aaronmdjones (approved):

PR #263, opened 2021-08-15
Stop announcing SASL cap changes on services split/join

glguy:
The current behavior creates a huge load on netjoin as thousands of already authenticated clients all try to reauthenticate at once for no particular benefit.

### DIFF
--- a/modules/m_sasl.c
+++ b/modules/m_sasl.c
@@ -53,13 +53,9 @@ static void abort_sasl(struct Client *);
 static void abort_sasl_exit(hook_data_client_exit *);
 
 static void advertise_sasl_cap(bool);
-static void advertise_sasl_new(struct Client *);
-static void advertise_sasl_exit(void *);
-static void advertise_sasl_config(void *);
 
 static unsigned int CLICAP_SASL = 0;
 static char mechlist_buf[BUFSIZE];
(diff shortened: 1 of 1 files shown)


dwfreed (changes requested):
As I said on IRC, I would rather see us simply reject attempts to reauth to solve your problem.  The 3.1 spec says that reauth ***MUST NOT*** be allowed, and the 3.2 spec says that reauth ***SHOULD*** be allowed.  But the 3.2 spec also says that we ***MUST NOT*** present the sasl cap if the authentication system is not available.  Thus, in 3.2 land, rejecting reauth would only break a should, but presenting sasl always would break a must.

jesopo:
I'd like to find out why it's a **MUST** because it's not a good idea. with my current understanding I'd be happy to break that piece of spec because it appears to be nonsense

kaniini:
I was apparently the person who wrote the *MUST* requirement, but I don't remember why.  I think it was probably politics.  It makes sense to downgrade it to *MAY* in my opinion.

dwfreed (approved):
After further discussion, my level of care for this has gone down to "eh, whatever."  Adding a ratelimited cap notify when the sasl agent returns would be really nice, but that can be done in a separate PR.

PR #265, opened 2021-08-19
hook_fn casts were hiding UB

glguy:


modules/m_alias.c
```diff
@@ -100,12 +100,11 @@ _moddeinit(void)
 	destroy_aliases();
 }
 
-static int
-reload_aliases(hook_data *data)
+static void
```
glguy, line 103: While the other cases were casting arguments this one was also casting a return type

dwfreed (approved):

PR #273, opened 2021-09-02
add gateway name field for auth blocks

### LINKED ISSUE #213: add "gateway name" field set by auth blocks and an extban that matches it
so, for example, someone using the libera webchat would have "is using gateway: web.libera.chat" in their WHOIS, and would match $g:web.libera.chat

examknow:
Closes #213. I'm marking this as a draft pending a response to https://github.com/solanum-ircd/solanum/issues/213#issuecomment-867985784.

### DIFF
--- a/extensions/cap_gateway.c
+++ b/extensions/cap_gateway.c
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (C) 2021 David Schultz <me@zpld.me>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307
+ *  USA
+ */
+
+#include <stdinc.h>
+#include <modules.h>
+#include <capability.h>
+#include <s_serv.h>
+#include <s_newconf.h>
+#include <client.h>
+#include <msgbuf.h>
+
+static char cap_gateway_desc[] = "Provides the solanum.chat/gateway capability";
+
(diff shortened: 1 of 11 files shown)


extensions/cap_gateway.c
```diff
@@ -0,0 +1,58 @@
...
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307
+ *  USA
+ */
+
+#include <stdinc.h>
+#include <modules.h>
+#include <capability.h>
+#include <s_serv.h>
+#include <s_newconf.h>
```
phy1729, line 24: I don't think these are necessary (MAPI_CAP_CLIENT is defined in modules.h).

extensions/extb_gateway.c
```diff
@@ -0,0 +1,54 @@
...
+	extban_table['w'] = NULL;
+}
+
+static int eb_gateway(const char *data, struct Client *client_p,
+		struct Channel *chptr, long mode_type)
+{
+	/* $w by itself will match all gateway users */
+	if (data == NULL)
+		return EmptyString(client_p->gateway) ? EXTBAN_NOMATCH : EXTBAN_MATCH;
+	return match(data, client_p->gateway) ? EXTBAN_MATCH : EXTBAN_NOMATCH;
```
phy1729, line 53: `match` seems to expect the second argument to be non `NULL`. I think this should be something like
```c
	if EmptyString(client_p->gateway)
		return EXTBAN_NOMATCH;
	/* $w by itself will match all gateway users */
	if (data == NULL)
		return EXTBAN_MATCH;
	return match(data, client_p->gateway) ? EXTBAN_MATCH : EXTBAN_NOMATCH;
```

modules/m_who.c
```diff
@@ -565,6 +568,21 @@ do_who(struct Client *source_p, struct Client *target_p, struct membership *mspt
 				q = "0";
 			append_format(str, sizeof str, &pos, " %s", q);
 		}
+		if (fmt->fields & FIELD_GATEWAY)
+		{
+			/* use same the logic as account */
```
phy1729, line 573: Given the reasoning for account to do this [1], I don't think it makes sense to copy that logic for gateway.

[1]: https://github.com/solanum-ircd/solanum/blob/0ca18d072a7d1dfcfbdff6dcf4f2788f52935a03/modules/m_services.c#L348-L350

phy1729 (commented):
A couple of places could add a `const` (`cap_gateway_desc`, `struct Client`'s `gateway`, and `struct ConfItem`'s `gateway`), but there doesn't seem to be a consistent style either way.

PR #275, opened 2021-09-06
Make opers talking through +g controllable by a user mode

spb:


modules/um_callerid.c
```diff
@@ -63,6 +63,16 @@ um_callerid_modinit(void)
 		return -1;
 	}
 
+	user_modes['@'] = find_umode_slot();
```
dwfreed, line 66: Can we... use something other than `@` for this mode? There are 52 alphabetical umodes, surely we haven't used all of them already.

modules/um_callerid.c
```diff
@@ -227,7 +239,34 @@ h_hdl_privmsg_user(void *vdata)
 	data->approved = ERR_TARGUMODEG;
 }
 
+static void
+check_umode_change(void *vdata)
+{
+	hook_data_umode_changed *data = (hook_data_umode_changed *)vdata;
+	bool changed = false;
```
dwfreed, line 246: "Variable set but not used" warning from GCC

modules/um_callerid.c
```diff
@@ -227,7 +239,34 @@ h_hdl_privmsg_user(void *vdata)
...
+	if (data->oldumodes & UMODE_OPER && !IsOper(source_p))
+		source_p->umodes &= ~user_modes['@'];
+
+	changed = ((data->oldumodes ^ source_p->umodes) & user_modes['@']);
+
+	if (source_p->umodes & user_modes['@'])
+	{
+		if (!HasPrivilege(source_p, "oper:message"))
+		{
+			sendto_one_notice(source_p, ":*** You need oper:message privilege for +p");
```
dwfreed, line 261: Mismatched umode

dwfreed (changes requested):

modules/um_callerid.c
```diff
@@ -227,7 +239,34 @@ h_hdl_privmsg_user(void *vdata)
...
+
+	if (!MyClient(source_p))
+		return;
+
+	if (data->oldumodes & UMODE_OPER && !IsOper(source_p))
+		source_p->umodes &= ~user_modes['@'];
+
+	changed = ((data->oldumodes ^ source_p->umodes) & user_modes['@']);
+
+	if (source_p->umodes & user_modes['@'])
```
glguy, line 257: This should be testing for the transition from `-@` to `+@` specifically. Presumably some aspect of `changed` was intended to handle this.

modules/um_callerid.c
spb: Yes. I figured there was going to be bikeshedding about the mode character whatever I used, so went with something I know everyone will hate to make sure it happens early.

modules/um_callerid.c
edk0: I think that's wrong; if you lose the oper:message privilege you should get +@ taken away too

modules/um_callerid.c
```diff
@@ -105,7 +117,7 @@ allow_message(struct Client *source_p, struct Client *target_p)
 		return true;
 
 	/* XXX: controversial?  allow opers to send through +g */
-	if (MayHavePrivilege(source_p, "oper:message"))
+	if (IsSetTalkThroughCallerID(source_p))
```
examknow, line 120: This makes it impossible for services to speak to a +g user without hacking the services code.

jesopo, line 120: does this prevent services messaging a `+g` user?

modules/um_callerid.c
```diff
@@ -227,7 +239,54 @@ h_hdl_privmsg_user(void *vdata)
...
+	if (data->oldumodes & UMODE_OPER && !IsOper(source_p))
+		source_p->umodes &= ~user_modes['M'];
+
+	changed = ((data->oldumodes ^ source_p->umodes) & user_modes['M']);
+
+	if (changed && source_p->umodes & user_modes['M'])
+	{
+		if (!HasPrivilege(source_p, "oper:message"))
+		{
+			sendto_one_notice(source_p, ":*** You need oper:message privilege for +M");
```
jesopo, line 261: could we also have this mode timeout like override does?

jesopo (commented):
some points from internal discussion

modules/um_callerid.c
glguy: I'd rather timeout behaviors be implemented client side if desired. Timeouts at best lead to ugly hacks in bots

jesopo:
is this waiting on anything

examknow:
https://github.com/solanum-ircd/solanum/blob/c77e470ee981bb57fed1af8465e0ed64a6b29b03/modules/um_callerid.c#L259

that should be `MayHavePrivilege`

dwfreed:
> https://github.com/solanum-ircd/solanum/blob/c77e470ee981bb57fed1af8465e0ed64a6b29b03/modules/um_callerid.c#L259
> 
> that should be `MayHavePrivilege`

That only applies to remote clients; the hook tests that the client is local before this.

dwfreed (approved):

PR #277, opened 2021-09-11
"KLINE active for ..." -> "Disconnected K-Lined user ..."

jesopo:
I've never liked the wording of this snote and it's lack of clear meaning confused someone today


edk0 (approved):

jillest:
There are also "DLINE active" and "XLINE active".

Apart from that, it's probably best to reword many server notices in one release rather than trickling such changes in over a long time, since scripts/bots may depend on the exact wording.

PR #278, opened 2021-09-15
Add support for bot mode spec

### LINKED ISSUE #210: Support umode +B for bots
unreal and related ircds support +B as an opt-in usermode. This is highly useful for bots that wish to automatically ignore other bots. Please consider supporting this, or an equivalent mechanism. Thank you.

examknow:
Closes #210

Link to spec for reference: https://ircv3.net/specs/extensions/bot-mode

### DIFF
--- a/extensions/Makefile.am
+++ b/extensions/Makefile.am
@@ -45,6 +45,7 @@ extension_LTLIBRARIES =		\
   sno_globaloper.la		\
   umode_noctcp.la		\
   m_adminwall.la		\
+  m_botmode.la			\
   m_echotags.la			\
   m_extendchans.la		\
   m_findforwards.la		\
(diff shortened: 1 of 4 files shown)


edk0:
I don't want draft caps. Independently of that I'm far from convinced that the bot mode is a good feature.

jesopo:
I'm also unconvinced this is the correct solution. I'd be a lot more interested in implementing the WIP `metadata` spec in anticipation of using that to denote bottiness

jesopo:
@TheDaemoness how do we feel about this these days

TheDaemoness:
(Pinging me because I've been a proponent for bot-mode on Libera staff.)

Seeing as `metadata` fell apart and Weechat now supports bot mode, I am generally in favor of implementing it, with perhaps the caveat that it should be possible for services to set this mode on users to facilitate things like bot accounts.

Also, this PR appears to be outdated: it calls for a `draft/bot` capability that is not present in [the final spec](https://ircv3.net/specs/extensions/bot-mode).

PR #281, opened 2021-09-15
Allow oper override to set forwards anywhere

### LINKED ISSUE #61: Extend oper:override to cover cmode +f
The user shouldn't need to be a member nor operator in either channel.

examknow:
Closes #61 - applies [suggestions](https://github.com/solanum-ircd/solanum/pull/117#issuecomment-771026111) from #117

### DIFF
--- a/extensions/override.c
+++ b/extensions/override.c
@@ -28,6 +28,7 @@ static const char override_desc[] =
 static void check_umode_change(void *data);
 static void hack_channel_access(void *data);
 static void hack_can_join(void *data);
+static void hack_can_forward(void *data);
 static void hack_can_kick(void *data);
 static void hack_can_send(void *data);
 static void hack_can_invite(void *data);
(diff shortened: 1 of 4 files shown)


edk0:
Thanks for the PR. So there's two reasons I suggested changing how `get_channel_access` is called. The first is that it's a step on the way to allowing override to work on channels you're not in in general, which might be a useful feature. But the one that's more directly relevant to this PR relates to module interaction. I think override should be aiming to report only if (a) the action was only allowed because of override, and (b) the action is going to go ahead. Using CHFL_OVERRIDE, while a bit of a hack, enables that quite nicely, since a high-priority hook (and honestly most of override should probably be using MONITOR rather than HIGHEST) can tell whether override was used (based on the alevel) and if the action is actually going to happen (based on the hook approval). While the rest of override is not doing this yet, that's the direction I think we should be going in.

PR #284, opened 2021-09-22
extb_canjoin: Ignore exemptions

edk0:


jesopo (approved):
comments above `is_banned()` and `is_banned_list()` definitions are notably wrong, but that's tangential to this PR

jesopo:
a note for the future is maybe it's a better idea to bitwise AND `CHFL_BAN` and `CHFL_EXCEPTION` when someone is +b **and** +e so you can differentiate "excepted" and "banned but excepted"

PR #285, opened 2021-09-23
Send RPL_ENDOFTRACE on local /etrace

skizzerz:
When running `/etrace <nick>` against a local user, the "end of trace" numeric is not sent to the client. This numeric is sent in all other circumstances (such as when running `/etrace <nick>` against a remote user), so send it when executing locally as well for consistency.


jesopo (approved):
same logic as `TRACE` now too

jesopo:
```
ETRACE jess
:sandcat.local 708 jess Oper users jess ~j localhost 127.0.0.1 j j :j
:sandcat.local 262 jess jess :End of TRACE
```
👌

PR #288, opened 2021-10-10
replace chm_staff and chm_hidden with cmode->priv and CHM_HIDDEN

jesopo:
the first commit in this returns a `*ChannelMode` from `cflag_add()` so that `->priv` and `->flags` can be modified without having to add a bunch of params to `cflag_add()`

the purpose for this was to give `+M`/operpeace a priv instead of the setter having to be an admin, but this also allowed me to get rid off the duplicated code in `chm_staff()` and `chm_hidden()`

### DIFF
--- a/extensions/chm_adminonly.c
+++ b/extensions/chm_adminonly.c
@@ -25,9 +25,11 @@ static struct ChannelMode *mymode;
 static int
 _modinit(void)
 {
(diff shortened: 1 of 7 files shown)


jesopo:
draft because I've realised `set_channel_mode` needs work to not send new-flavour hidden modes to everyone

PR #290, opened 2021-10-16
m_sasl: Remove implicit abort on registration

edk0:
This doesn't make sense in a world where post-registration SASL is allowed, and should fix one case of an annoying login desync that's seen in the real world.

Specifically, when a client sends its final AUTHENTICATE and Atheme receives it, it sends an SVSLOGIN for that client. If the client sends us its CAP END *before* we see the SVSLOGIN, the implicit abort will try to abort the SASL session that's already succeeded.

Atheme interprets this as an instruction to forget about the successful SASL session; you'll connect unidentified. But it's already sent SVSLOGIN, which will log the client in ircd-side, causing ircd and services views to differ until the user authenticates again manually.

I think allowing a SASL session to be aborted when it has already succeeded is an Atheme bug, and it can still be triggered without this change. But our behaviour here seems silly anyway.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,8 @@ bolded warnings in the full release notes below.
 - Add `--with-asan` to configure to produce an ASan instrumented build
 
 ### server protocol
+- **Breaking:** Don't implicitly abort SASL when connection registration handshake completes;
+  requires updating atheme to include https://github.com/atheme/atheme/pull/833.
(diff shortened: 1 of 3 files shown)


dwfreed (changes requested):
This just needs a NEWS addition noting that it needs Atheme with the referenced PR merged and then this can be merged.

dwfreed (approved):

dwfreed (approved):

PR #302, opened 2021-12-04
channel: always send chm_hidden modes to servers

edk0:


jesopo (approved):
`channel_modes()` is called for both `RPL_CHANNELMODEIS` (to clients) and `SJOIN` (to servers), and `auspex:cmodes` only matters if you're a client, and `client_p` is not a client when we're here due to `SJOIN`. LGTM

jesopo:
though I'm not exactly sure how the code ended up wrong in the first place 😅

PR #303, opened 2021-12-04
snote when an oper sets a chm_hidden mode

### LINKED ISSUE #105: snote when an oper sets/unsets a chm_hidden mode
setting/unsetting a `chm_hidden` mode feels like a bit of an event given the only one we've got is `chm_operpeace`, so I think it makes sense to let everyone know

jesopo:
unideal for code duplication. open to better suggestions

closes #105

### DIFF
--- a/ircd/chmode.c
+++ b/ircd/chmode.c
@@ -1538,8 +1538,13 @@ set_channel_mode(struct Client *client_p, struct Client *source_p,
 				*mbuf = '\0';
 
 				if(cur_len > mlen)
+				{
 					sendto_channel_local_priv(IsServer(source_p) ? fakesource_p : source_p,
 							send_flags, priv, chptr, "%s %s", modebuf, parabuf);
+					if (flags == ONLY_OPERS && IsClient(source_p))
+						sendto_realops_snomask(SNO_GENERAL, L_ALL, "%s set hidden modes on %s: %s",
+							get_oper_name(source_p), chptr->chname, &modebuf[mlen]);
+				}
 				else
 					continue;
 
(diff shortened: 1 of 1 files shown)


jesopo:
maybe this should go for oper-only but non-hidden modes too like `+LP`

ircd/chmode.c
```diff
@@ -1538,8 +1538,13 @@ set_channel_mode(struct Client *client_p, struct Client *source_p,
 				*mbuf = '\0';
 
 				if(cur_len > mlen)
+				{
 					sendto_channel_local_priv(IsServer(source_p) ? fakesource_p : source_p,
 							send_flags, priv, chptr, "%s %s", modebuf, parabuf);
+					if (priv != NULL && IsClient(source_p))
```
dwfreed, line 1544: This doesn't appear to be the test you think it is? `priv` is only set to non-null when `flags` is `ONLY_OPERS` (in which case it is set to `auspex:cmodes`, which is not the priv used to set the mode), and the only mode changes with `mems` set to `ONLY_OPERS` are `chm_hidden` modes.

ircd/chmode.c
```diff
@@ -1575,8 +1580,14 @@ set_channel_mode(struct Client *client_p, struct Client *source_p,
 
 		*mbuf = '\0';
 		if(cur_len > mlen)
+		{
 			sendto_channel_local_priv(IsServer(source_p) ? fakesource_p : source_p,
 				send_flags, priv, chptr, "%s %s", modebuf, parabuf);
+			if (priv != NULL && IsClient(source_p))
```
dwfreed, line 1586: Ditto

dwfreed (changes requested):

PR #306, opened 2022-01-01
move privset extban to $O and add $o extban for oper name

jesopo:


jesopo:
I'd prefer to change whatever is casefolding it instead of picking a different extban char. `o` vs `O` was my choice given the `STATS` chars

jesopo:
zoop :point_right: #307

PR #307, opened 2022-01-01
don't casefold extban characters

jesopo:


jillest (commented):
This is not sufficient. There is another `extban_table[(unsigned char) irctolower(*p)]` below and there is also code that normalizes the extban character to lowercase when setting a ban. Perhaps this normalization should be skipped only if the uppercase extban exists, to avoid breaking existing behaviour.

Also, the uppercase and lowercase extbans need to know about each other since it's not possible to have two bans in a channel that only differ in case. Some clients such as irssi track ban lists so this may well be a client restriction as well.

jesopo:
I don't mind about the client angle given other popular IRCd software does this

jesopo:
> I don't mind about the client angle given other popular IRCd software does this

(though I'm willing to lose this debate)

jillest:
It seems like it would be best to make only the extban character case sensitive, so users manually typing in ban masks for `-b` get surprising results as infrequently as possible. A case sensitive extban character would be like modes.

Another alternative might be allowing extbans to be specified by whole words. I guess this was originally not done to avoid having too many ;-)

jesopo:
> It seems like it would be best to make only the extban character case sensitive, so users manually typing in ban masks for `-b` get surprising results as infrequently as possible. A case sensitive extban character would be like modes.

is this not what the code currently does?

PR #308, opened 2022-01-01
send RPL_LOGGEDIN to target_p upon ENCAP SU

jesopo:
intended usecase: a client receives `RPL_LOGGEDIN` post-registration on a network that supports SASL but the client knows it doesn't currently have a SASL configuration, so it prompts the user with "would you like to save your account password for this network?"


A-UNDERSCORE-D (dismissed):

edk0:
This sends a RPL_LOGGEDIN for `(null)` if I log out. It should probably send RPL_LOGGEDOUT or nothing

dwfreed (changes requested):
What @edk0 said; I prefer the RPL_LOGGEDOUT response.

dwfreed (approved):

PR #310, opened 2022-01-12
don't subject remote users to chm_nonotice

jesopo:
chanserv currently cannot notice `VERBOSE` channels that are `+T` and it feels needless to check this on every single server of the network


aaronmdjones (dismissed):
/me grumbles something about the comment above it being inaccurate

jesopo:
better?

dwfreed (approved):

PR #319, opened 2022-03-18
tests/misc.c: use a duration which won't overflow a 31-bit integer

aaronmdjones:
`7000` weeks is `4233600000` seconds, which overflows a 31-bit integer. This causes spurious test failures on machines where a (signed) `time_t` is 32 bits.


dwfreed (approved):

aaronmdjones:
```
<edk> can we please not merge #319
<edk> i'm not signed into github here so i can't press teh buttons to block it
<edk> but i have a change in the works to fix it by handling overflow correctly
<edk> i have explicit overflow checks before each operation, if we overflow we just return early with the maximum time
```

PR #323, opened 2022-04-01
whowas.c: store account name in whowas

examknow:


glguy:
Is it ever possible for `client_p->user` to be `NULL` at this point? (It seems likely that `whowas_add_history` is only ever called on such `client_p`s but I'm just asking to be sure)

examknow:
> Is it ever possible for `client_p->user` to be `NULL` at this point? (It seems likely that `whowas_add_history` is only ever called on such `client_p`s but I'm just asking to be sure)

I had originally had put an `IsPerson()` check but after looking around I couldn't find any circumstance where `client_p->user` would be `NULL` so I ended up removing it.

edk0 (approved):
LGTM.

strcpy convention aside, this is the exact implementation that was introduced in 01b7a527a37e111d6e2e1b9427ed0094fd959893 and subsequently lost

PR #325, opened 2022-04-01
Remove the subset ban restriction

### LINKED ISSUE #321: *!*@1.2.3.4/24 is seen as a subset of *!*@*/*
solanum refuses to set bans that it thinks affects a subset of something already banned by a superset, and it incorrectly thinks a CIDR ban matches a glob ban

### LINKED ISSUE #322: no error returned for bans that refuse to be set because they're a subset of an already banned superset
this makes sense to a degree; the existing behaviour for redundant mode changes is to just do nothing, but this is confusing when a ban refuses to set because it's a subset of an already banned superset, because it's not always easy to figure out what superset ban exists (see also #321, where the superset was actually not a superset)

edk0:
I think this was always pretty questionable. You can set redundant bans in various ways anyway, and preventing all of them would only make the situation worse, as wide temporary bans would destroy narrow permanent ones, for example.

Fixes #321, fixes #322

### DIFF
--- a/ircd/chmode.c
+++ b/ircd/chmode.c
@@ -249,9 +249,7 @@ add_id(struct Client *source_p, struct Channel *chptr, const char *banid, const 
 	char *realban = LOCAL_COPY(banid);
 	rb_dlink_node *ptr;
 
-	/* dont let local clients overflow the banlist, or set redundant
-	 * bans
-	 */
+	/* dont let local clients overflow the banlist */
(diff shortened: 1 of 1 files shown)


glguy (approved):
I like this change.

I think it's common to want a redundant ban especially while refining a general ban before you've removed the general ban. Furthermore the feedback and which ban was overlapping was missing which could make it especially confusing.

PR #326, opened 2022-04-08
Remove ERR_NOSUCHNICK from LIST reply

progval:
1. The first block was unreachable according to @dwfreed
2. In the second block, sending ERR_NOSUCHNICK can be unexpected by clients,
   as no other IRCd does that and https://modern.ircdocs.horse/#list-message
   does not list it.
   Additionally, it leaks information on the existence of secret channels.

### DIFF
--- a/modules/m_list.c
+++ b/modules/m_list.c
@@ -436,18 +436,10 @@ static void safelist_channel_named(struct Client *source_p, const char *name, in
 	if ((p = strchr(name, ',')))
 		*p = '\0';
 
-	if (*name == '\0')
-	{
-		sendto_one_numeric(source_p, ERR_NOSUCHNICK, form_str(ERR_NOSUCHNICK), name);
-		sendto_one(source_p, form_str(RPL_LISTEND), me.name, source_p->name);
-		return;
-	}
-
 	chptr = find_channel(name);
 
 	if (chptr == NULL)
 	{
-		sendto_one_numeric(source_p, ERR_NOSUCHNICK, form_str(ERR_NOSUCHNICK), name);
 		sendto_one(source_p, form_str(RPL_LISTEND), me.name, source_p->name);
 		return;
 	}


edk0:
"Leaking" information on the existence of secret channels is expected and unavoidable. If you know a secret channel's name you can just try to join it...

I think sending the 401 is useful and I don't believe we should stop just because some documentation somewhere doesn't describe this behaviour. If there's a functionally equivalent standard solution we can switch to it, if not the standard should expand.

progval:
The "If there's a functionally equivalent standard solution" is to omit the channel from the output

edk0:
I'm not interested in playing word games. I'm not going to remove a feature just because you don't like it and Dan didn't know about it.

progval:
What word games?

I'm saying `LIST #chan` returns this if the channel exists:

```
:server 321 bar Channel :Users  Name
:server 322 bar #chan 1 :channel topic
:server 323 bar :End of /LIST
```

and this if it doesn't:

```
:server 321 bar Channel :Users  Name
:server 323 bar :End of /LIST
```

Literally everyone but Solanum (and Charybdis) do this, that's why I call it a standard solution. (Though Ergo currently returns ERR_NOSUCHCHANNEL, but it [causes issues with Gamja](https://github.com/ergochat/ergo/issues/1928), so they will remove it in the next release)

jesopo:
what's the benefit of changing this behaviour

progval:
To make IRC less fragmented.

Clients may not expect it to be a reply to `LIST` and show a popup or something.

It's also inconsistent with `WHO`, as Solanum does not reply `ERR_NOSUCHNICK` when given a non-existent nick as argument.

jesopo:
do you have any feasible issue caused by this specific inconsistency

progval:
https://github.com/ergochat/ergo/issues/1928 is a close one.

It just doesn't happen because Gamja sends `LIST #idontexist*` instead of `LIST #idontexist`, in which case Solanum omits the numeric.

jesopo:
ok. my concern is it's easier for a human and a bot to understand "doesn't exist" than it does having the numeric omitted

progval:
Then why doesn't solanum send ERR_NOSUCHNICK for WHO or NAMES?

And it doesn't matter for bots, because they have to handle servers which don't send ERR_NOSUCHNICK as a reply to LIST anyway.

jesopo:
well, no they don't. i only write bots for solanum. i don't think we should pick the less good option here just because the less good option is more common

edk0:
I consider the error message when you `LIST #nonexistent` to be a feature; it's definitely much easier for a human to recognise that "#nonexistent: no such nick/channel" is about #nonexistent as opposed to "End of /LIST". Is it perfect? Well, no, and you've identified some of the ways in which it is not. But I don't think any of them is so bad that deleting the feature would be an improvement.

I'm saying I'd be OK with improving the behaviour here if there were a standard way to implement the feature. Not implementing the feature doesn't meet my requirements.

progval:
What about writing "No such channel" in the trailing parameter of RPL_LISTEND?

emersion:
Maybe I can explain my use-case a bit better. My issues with Solanum's behavior stem from two things:

- My client surfaces all error messages to the user. When receiving `ERR_NOSUCHNICK`, an error popup will appear with the message. Responses from the server can't reliably be linked to commands sent by the client.
- My client has a text field to join a channel. The client will send a `LIST` command to display the topic of the currently typed channel name. The `LIST` command will be sent when the user pauses typing for 300ms. This allows the user to get feedback before clicking "Join". If the user does a slight pause while typing (e.g. means to type `#awesome-channel` but stops a bit at `#awesome-`) then Solanum sends an error message which gets displayed right away.

This is a bit confusing for users and I can't think of a good way to avoid this. I'd prefer not to add a Solanum-specific workaround.

PR #327, opened 2022-04-13
Add description parameter to auth blocks

glguy:
This would make it easy to see _why_ we have so many I lines and to group them back together to a single auth block both with manual queries and also for consumption by automated tooling.

### DIFF
--- a/modules/m_testline.c
+++ b/modules/m_testline.c
@@ -34,7 +34,6 @@
 #include "msg.h"
 #include "hash.h"
 #include "hostmask.h"
-#include "match.h"
 #include "numeric.h"
 #include "s_conf.h"
 #include "s_newconf.h"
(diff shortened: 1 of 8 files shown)


modules/m_testline.c
```diff
@@ -235,7 +235,8 @@ mo_testline(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *sou
 		sendto_one_numeric(source_p, RPL_STATSILINE, form_str(RPL_STATSILINE),
 				aconf->info.name, EmptyString(aconf->spasswd) ? "<NULL>" : aconf->spasswd,
 				show_iline_prefix(source_p, aconf, aconf->user),
-				aconf->host, aconf->port, aconf->className);
+				aconf->host, aconf->port, aconf->className,
+				EmptyString(aconf->desc) ? "" : aconf->desc);
```
dwfreed, line 239: Why not reuse `CheckEmpty()` here?

glguy: First because I didn't know about it at first and then second because it wasn't already in scope and I was avoiding adding new includes, but it's probably better to just use it.

dwfreed (approved):

PR #328, opened 2022-04-14
Apply cmode C to replies

glguy:


modules/chm_noctcp.c
```diff
@@ -51,10 +51,12 @@ chm_noctcp_process(void *data_)
...
 	hook_data_privmsg_channel *data = data_;
 	/* don't waste CPU if message is already blocked */
-	if (data->approved || data->msgtype == MESSAGE_TYPE_NOTICE)
+	if (data->approved)
 		return;
 
-	if (*data->text == '\001' && rb_strncasecmp(data->text + 1, "ACTION ", 7) && data->chptr->mode.mode & mode_noctcp)
+	if (*data->text == '\001' &&
+	    data->chptr->mode.mode & mode_noctcp &&
+	    !(data->msgtype == MESSAGE_TYPE_PRIVMSG && !rb_strncasecmp(data->text + 1, "ACTION ", 7)))
```
jesopo, line 59:
```C
(data->msgtype == MESSAGE_TYPE_NOTICE || !rb_strncasecmp(data->text + 1, "ACTION ", 7))
```
this seems easier to understand?

jesopo (commented):
no big deal, just took me a while to understand what it was doing

modules/chm_noctcp.c
glguy: I wanted to be explicit that we're not blocking things that are actions, rather than listing all the ways something might not be an action.

jesopo: fair enough.

jesopo (approved):

dwfreed (approved):

PR #329, opened 2022-04-15
refuse to honour RSFNC for a RESV nick

jesopo:


dwfreed:
This breaks keeping a RESV for `Guest*` so that the only way a person can have a guest nick is via services.  OFTC currently has such a RESV, though I suspect part of our reasoning is that our guest nicks are sequential and we do not have SAVE, so a determined attacker could kill a guesting client by colliding with their nick change.

jesopo:
we could solve that by having 2 kinds of RSFNC. one that is expected to sometimes fail and one that should never fail, and you'd use the former for any user-supplied nickname

dwfreed:
Atheme already allows differentiating between a regain and a force change via the `fnc_sts` handler; `FNC_REGAIN` is the type value for regain, and `FNC_FORCE` is the type value for force changes.  If we pursue that route, I think `RSFNC` should remain the `FNC_FORCE` handler, and we could add something like `ENCAP REGAIN` for the `FNC_REGAIN` handler, which would require compliance with all applicable nick rules, just like `NICK`

jesopo:
atheme work done here https://github.com/atheme/atheme/pull/864 so now we can think about how this should look over the wire

PR #330, opened 2022-04-15
run spamfilters against NICK

jesopo:
this isn't perfect because it'll format as
```
0:nick1!user@host#0 NICK :nick2
```
(note the `:`) so maybe I should beat up `match_message`'s `snprintf` a bit

### DIFF
--- a/extensions/filter.c
+++ b/extensions/filter.c
@@ -56,6 +56,7 @@ static const char filter_desc[] = "Filter messages using a precompiled Hyperscan
 static void filter_msg_user(void *data);
 static void filter_msg_channel(void *data);
 static void filter_client_quit(void *data);
+static void filter_client_nick(void *data);
 static void on_client_exit(void *data);
 
 static void mo_setfilter(struct MsgBuf *, struct Client *, struct Client *, int, const char **);
@@ -93,6 +94,7 @@ mapi_hfn_list_av1 filter_hfnlist[] = {
 	{ "privmsg_user", filter_msg_user },
 	{ "privmsg_channel", filter_msg_channel },
 	{ "client_quit", filter_client_quit },
+	{ "local_nick_change", filter_client_nick },
 	{ "client_exit", on_client_exit },
 	{ NULL, NULL }
 };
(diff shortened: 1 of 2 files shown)


dwfreed:
This should also apply to the initial nick, otherwise one could still register with the bad nick. Alarm consumers would need to handle that situation, though.

extensions/filter.c
```diff
@@ -483,6 +489,54 @@ filter_client_quit(void *data_)
...
+			NULL, NULL, NULL, addr, CONF_EXEMPTDLINE | 1, GET_SS_FAMILY(addr), NULL, NULL))
+		return;
+
+	unsigned r = match_message("0", NULL, "NICK", NULL, data->nick);
+	if (r & ACT_DROP) {
+		data->approved = 0;
+	}
+	if (r & ACT_ALARM) {
+		sendto_realops_snomask(SNO_GENERAL, L_ALL | L_NETWIDE,
+			"FILTER:REGISTER: %s@%s",
```
jesopo, line 532: this is my initial solution for how alarm consumers are going to handle this case: slightly different snote. bikeshed on the format please

extensions/filter.c
```diff
@@ -483,6 +489,54 @@ filter_client_quit(void *data_)
...
+}
+
+void
+filter_client_nick_set(void *data_)
+{
+	hook_data_nick_approval *data = data_;
+	struct Client *s = data->client;
+	struct sockaddr *addr = (void *)&s->localClient->ip;
+
+	if(find_conf_by_address(
```
jesopo, line 522: we can't make people immune from this by them being an oper, obviously, and this is the only alternative that came to mind

jesopo: or maybe we just treat it like a RESV. you have to connect first and then switch your nick

jesopo:
i plan to do a second PR after this that runs spamfilters against `USER`; when i do this im going to move away from using `match_message(...)` for everything and refactor out the bits of `match_message(...)` that are useful for things that are not messages, but i thought i'd keep this simple to start with.

PR #335, opened 2022-05-17
ERROR instead of NOTICE for failed WEBIRC

jesopo:
per the [IRCv3 WEBIRC](https://ircv3.net/specs/extensions/webirc) spec:

> If the connection fails to the IRC server, the WebIRC gateway MUST return an ERROR with an appropriate message and terminate the connection.


dwfreed (changes requested):
These should just use `exit_client`, which will do the `ERROR` and proper closing of the connection for you

dwfreed (approved):

PR #336, opened 2022-05-21
Allow for default umodes to change at runtime

skizzerz:
When default umodes are changed in ircd.conf, these changes will now be
applied to all local clients upon receiving a /rehash. We track which
umodes a user has explicitly (un)set and avoid modifying those umodes
when changing the defaults -- only umodes that a user didn't explicitly
touch are subject to change on rehash.

### DIFF
--- a/include/client.h
+++ b/include/client.h
@@ -234,6 +234,9 @@ struct LocalUser
 	/* nicknames theyre monitoring */
 	rb_dlink_list monitor_list;
 
+	/* collects umodes explicitly changed by this user */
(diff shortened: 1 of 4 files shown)


edk0:
Hmm. It'd be nice if setting umode -R when you're _already_ -R would mark R as a changed umode, but clearly harder to do that with how user_mode works now.

skizzerz:
@edk0 Wasn't significantly harder to do, so I added that feature in as well.

edk0:
This currently sets people -Z when the default umodes change

PR #341, opened 2022-06-09
don't truncate operspy WHO

### LINKED ISSUE #340: don't truncate operspy WHO output

jesopo:
closes #340


jillest (approved):
Looking again, the other decrement is in `who_common_channel` which isn't used in the operspy case, but still this new approach is better because it's more obviously correct.

edk0 (approved):

PR #344, opened 2022-06-16
optionally allow WEBIRC blocks to spoof ident responses

jesopo:


dwfreed:
Note that authproc does not check whether `IsGotId` is set already, so if for some reason the webirc gateway is running both identd and wanting to spoof username via `WEBIRC` the identd will most likely override the spoof.

dwfreed (changes requested):

include/client.h
```diff
@@ -419,6 +419,7 @@ struct ListClient
 #define FLAGS_EXEMPTSHIDE	0x04000000
 #define FLAGS_EXEMPTJUPE	0x08000000
 #define FLAGS_IDENTIFIED	0x10000000	/* owns their current nick */
+#define FLAGS_USER_SPOOFING	0x00000080	/* username has already been set, so ignore it in USER */
```
dwfreed, line 422: Collides with `FLAGS_GOTID`

ircd/s_user.c
```diff
@@ -500,7 +500,7 @@ register_local_user(struct Client *client_p, struct Client *source_p)
 		}
 
 		/* dont replace username if its supposed to be spoofed --fl */
-		if(!IsConfDoSpoofIp(aconf) || !strchr(aconf->info.name, '@'))
+		if(!IsUserSpoof(source_p))
```
dwfreed, line 503: The above-mentioned collision means that this if is always true

dwfreed (changes requested):

PR #345, opened 2022-06-16
extensions/chm_serviceonly.c: only allow service users to join

jesopo:
intended usecase is to allow for channels that are IPC between atheme and a pseudoserver; this may be too niche for upstream

### DIFF
--- a/extensions/chm_serviceonly.c
+++ b/extensions/chm_serviceonly.c
@@ -1,0 +1,56 @@
+#include "stdinc.h"
+#include "modules.h"
+#include "hook.h"
+#include "client.h"
+#include "ircd.h"
+#include "send.h"
+#include "s_conf.h"
+#include "s_user.h"
+#include "s_serv.h"
+#include "numeric.h"
+#include "chmode.h"
+
+static const char chm_serviceonly_desc[] =
+	"Adds channel mode +X which makes a channel service-only";
+
+static void h_can_join(void *);
+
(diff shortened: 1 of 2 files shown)


extensions/chm_serviceonly.c
```diff
@@ -0,0 +1,56 @@
...
+	{ "can_join", h_can_join },
+	{ NULL, NULL }
+};
+
+static unsigned int mymode;
+
+static int
+_modinit(void)
+{
+	mymode = cflag_add('X', chm_staff);
```
jesopo, line 28: placeholder character, please suggest better options

extensions/chm_serviceonly.c
```diff
@@ -0,0 +1,56 @@
...
+DECLARE_MODULE_AV2(chm_serviceonly, _modinit, _moddeinit, NULL, NULL, serviceonly_hfnlist, NULL, NULL, chm_serviceonly_desc);
+
+static void
+h_can_join(void *data_)
+{
+	hook_data_channel *data = data_;
+	struct Client *source_p = data->client;
+	struct Channel *chptr = data->chptr;
+
+	if((chptr->mode.mode & mymode) && !IsService(source_p)) {
```
jillest, line 51: There is no way a local client can be a service, so the `&& !IsService(source_p)` part could be removed.

This feels weird.

TwinUsers (approved):

PR #348, opened 2022-06-27
Honor +I exemptions for +k

uplime:
Allow users to bypass a key requirement from `+k` if they have an invite exemption.

### DIFF
--- a/ircd/channel.c
+++ b/ircd/channel.c
@@ -716,7 +716,22 @@ can_join(struct Client *source_p, struct Channel *chptr, const char *key, const 
 		goto finish_join_check;
 	}
 
-	if(*chptr->mode.key && (EmptyString(key) || irccmp(chptr->mode.key, key)))
+	struct Ban *invex = NULL;
+	struct matchset ms;
+	rb_dlink_node *ptr;
+
+	matchset_for_client(source_p, &ms);
+
+	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
+	{
+		invex = ptr->data;
+		if (matches_mask(&ms, invex->banstr) ||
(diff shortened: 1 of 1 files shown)


ircd/channel.c
```diff
@@ -716,12 +716,25 @@ can_join(struct Client *source_p, struct Channel *chptr, const char *key, const
...
+
+	RB_DLINK_FOREACH(ptr, chptr->invexlist.head)
+	{
+		invex = ptr->data;
+		if (matches_mask(&ms, invex->banstr) ||
+				match_extban(invex->banstr, source_p, chptr, CHFL_INVEX))
+			break;
+	}
+
+	if(*chptr->mode.key && !invex &&
```
dwfreed, line 729: `invex` will always be valid here, thus always allowing people in.  You need to `NULL` it at the end of every loop above, so that the break skips that and thus it's only valid if it matched.

jillest:
I'm not sure this exact change to how channel keys work is the right solution for your issues. Right now, we only deliberately expose the key to channel members (which must have initially known the key to get in, or seen it being set) and to opers with an explicit operspy command (/mode !#channel); apart from that, split abusers might also get the key, and ChanServ has a GETKEY command.

Perhaps this kind of override works better if only channel operators can see keys, so people who can get in via +I don't get the key.

Also, it seems reasonable for me that /invite overrides as least as much as a +I entry.

uplime:
That isn't something I had considered. However for invite exemptions, it doesn't seem as problematic as a split abuser (or anything similar), since this would mean somebody with authority to created the exemption. For what its worth, my use-case isn't to secure the channel key, but to ensure a trusted regular has access to such a channel but might not be connecting from the same client every time (if they have to move between multiple computers for whatever reason) and don't or aren't interested in using a bouncer.

PR #349, opened 2022-06-28
chmode: Use original string when removing a mask

phy1729:
With this
```
MODE #test +b foo!*@*
MODE #test -b Foo!*@*
```
outputs `MODE #test -b foo!*@*` rather than `MODE #test -b Foo!*@*` which makes state tracking easier.


edk0 (changes requested):
I'd like to think about this more

dwfreed (approved):

edk0 (approved):

PR #350, opened 2022-07-01
Cast time_t to long long when printing

phy1729:
In particular OpenBSD uses `long long` for `time_t`. Could instead cast to `intmax_t` though I doubt that'd differ in practice.


dwfreed (approved):
Also fixes the instances where `time_t` is larger than `long` on Linux (this is possible with recent glibc versions on x86)

PR #354, opened 2022-07-17
EBMASK capab, to burst BMASK metadata

jesopo:


ircd/s_serv.c
```diff
@@ -510,26 +512,37 @@ burst_modes_TS6(struct Client *client_p, struct Channel *chptr,
...
+			sprintf(bmask, "%s$%s", banptr->banstr, banptr->forward);
+		else
+			strcpy(bmask, banptr->banstr);
+
+		if (IsCapable(client_p, CAP_EBMASK))
+			sendto_one(client_p, "%s %s %ld %s",
+				buf,
+				bmask,
+				(long)banptr->when,
+				banptr->who);
```
jillest, line 539: Even though this will work, I think it is clearer to have a separate command (named EBMASK) with the new format. This avoids the need to check whether the other side thinks we have the EBMASK capab.

ircd/s_serv.c
```diff
@@ -510,26 +512,37 @@ burst_modes_TS6(struct Client *client_p, struct Channel *chptr,
...
 		else
-			send_multiline_item(client_p, "%s", banptr->banstr);
+			send_multiline_item(client_p, "%s", bmask);
 	}
 
-	send_multiline_fini(client_p, NULL);
+	if (!IsCapable(client_p, CAP_EBMASK))
+		send_multiline_fini(client_p, NULL);
```
jillest, line 545: Code-wise, it might be better to have a separate `burst_modes_EBMASK` and leave `burst_modes_TS6` as is, to avoid the complicated logic with multiple `CAP_EBMASK` checks. I think it is acceptable to iterate over the ban list twice; the optional `%s$%s` construction might be extracted to a function.

This applies particularly if it is planned to deprecate non-EBMASK entirely eventually since `burst_modes_TS6` can be removed at that point without untangling any logic.

ircd/s_serv.c
```diff
@@ -522,11 +525,14 @@ burst_modes_TS6(struct Client *client_p, struct Channel *chptr,
...
-			send_multiline_item(client_p, "%s$%s",
-					banptr->banstr,
-					banptr->forward);
+			sprintf(buf, "%s$%s", banptr->banstr, banptr->forward);
 		else
-			send_multiline_item(client_p, "%s", banptr->banstr);
+			strcpy(buf, banptr->banstr);
+
+		if IsCapable(client_p, CAP_EBMASK)
+			sprintf(buf, "%s,%ld,%s", buf, (long)banptr->when, banptr->who);
```
jillest, line 533: Making `sprintf()` copy between overlapping objects (`buf`) is UB. Perhaps it's better to use two `send_multiline_item` calls, one for traditional and one for EBMASK.

modules/core/m_mode.c
```diff
@@ -368,8 +371,23 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
...
+				bants = atol(strtok(NULL, " "));
+				who = strtok(NULL, " ");
+
+				if (who == NULL)
+				{
+					// die here, not enough params
+				}
+
+				banptr->when = bants;
+				strcpy(banptr->who, who);
```
dwfreed, line 388:
```suggestion
				rb_free(banptr->who);
				banptr->who = rb_strdup(who);
```

`who` in `struct Ban` is not a static size buffer, but the result of an `rb_strdup`; do the same here.

dwfreed (changes requested):
Seems sane otherwise

modules/core/m_mode.c
```diff
@@ -391,21 +409,14 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 			pbuf += arglen;
 			plen += arglen;
 			modecount++;
+			sprintf(dbuf, "%s ", s);
```
dwfreed, line 412: You aren't advancing `dbuf` after this, so you're simply overwriting the start of the buffer over and over again.

modules/core/m_mode.c
```diff
@@ -415,6 +426,21 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 		sendto_channel_local(fakesource_p, mems, chptr, "%s %s", modebuf, parabuf);
 	}
 
-	sendto_server(client_p, chptr, CAP_TS6 | needcap, NOCAPS, ":%s BMASK %ld %s %s :%s",
-		      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], parv[4]);
+	if (extended)
+		sendto_server(client_p, chptr, CAP_EBMASK | CAP_TS6 | needcap, NOCAPS, ":%s EBMASK %ld %s %s :%s",
+			      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], parv[4]);
+	sendto_server(client_p, chptr, CAP_TS6 | needcap, CAP_EBMASK, ":%s BMASK %ld %s %s :%s",
```
dwfreed, line 432: This won't send `BMASK` bans to `EBMASK`-capable servers (@edk0 gets the credit for this one)

dwfreed (changes requested):

modules/core/m_mode.c
```diff
@@ -327,29 +336,28 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
...
-	/* next char isnt a space, point t to the next one */
-	if((t = strchr(s, ' ')) != NULL)
-	{
-		*t++ = '\0';
-
-		/* double spaces will break the parser */
-		while(*t == ' ')
-			t++;
-	}
+	strtok(s, " ");
```
dwfreed, line 348:
```suggestion
	s = strtok(s, " ");
```

modules/core/m_mode.c
```diff
@@ -368,8 +376,32 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
...
+					snprintf(squitreason, sizeof squitreason, "Invalid EBMASK chunk");
+					exit_client(source_p, source_p, source_p, squitreason);
+					return;
+				}
+
+				banptr->when = bants;
+				rb_free(banptr->who);
+				banptr->who = rb_strdup(who);
+
+				sprintf(dbuf, "%s ", s);
```
dwfreed, line 401:
```suggestion
				arglen = sprintf(dbuf, "%s ", s);
```

dwfreed (changes requested):

modules/core/m_mode.c
dwfreed: Alternate suggestion (and then remove the next line, which I can't do in this suggestion): 

```suggestion
				dbuf += sprintf(dbuf, "%s ", s);
```

ircd/s_serv.c
```diff
@@ -522,11 +525,18 @@ burst_modes_TS6(struct Client *client_p, struct Channel *chptr,
...
+			sprintf(buf, "%s$%s", banptr->banstr, banptr->forward);
 		else
-			send_multiline_item(client_p, "%s", banptr->banstr);
+			strcpy(buf, banptr->banstr);
+
+		if IsCapable(client_p, CAP_EBMASK)
+			send_multiline_item(client_p, "%s %ld %s",
+				buf,
+				(long)banptr->when,
+				banptr->who);
```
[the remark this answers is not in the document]
jesopo: what's the alternative?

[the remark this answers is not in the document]
jillest: An alternative would be to keep bans that were set before EBMASK was introduced special forever, but I'm not sure it's worth it. Looking at functionality, a ban set by a server means the true setter is not known and the ban is at least as old as the given timestamp, and that remains the case.

Displaying the name of the server that was first upgraded to EBMASK is a violation of flatten_links, strictly speaking, but it's not very important.

modules/core/m_mode.c
```diff
@@ -368,53 +377,83 @@ ms_bmask(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
...
+		*(param_ptr - 1) = '\0';
+		*(degrade_ptr - 1) = '\0';
+		sendto_channel_local(fakesource_p, mems, chptr, "%s %s", output, parabuf);
+	}
+
+	if (extended) {
+		sendto_server(client_p, chptr, CAP_EBMASK | CAP_TS6 | needcap, NOCAPS, ":%s EBMASK %ld %s %s :%s",
+			      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], parv[4]);
+		sendto_server(client_p, chptr, CAP_TS6 | needcap, CAP_EBMASK, ":%s BMASK %ld %s %s :%s",
+			      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], degrade);
```
dwfreed, line 442:
```suggestion
		*output_ptr = '\0';
		*(param_ptr - 1) = '\0';
		sendto_channel_local(fakesource_p, mems, chptr, "%s %s", output, parabuf);
	}

	if (extended) {
		*(degrade_ptr - 1) = '\0';

		sendto_server(client_p, chptr, CAP_EBMASK | CAP_TS6 | needcap, NOCAPS, ":%s EBMASK %ld %s %s :%s",
			      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], parv[4]);
		sendto_server(client_p, chptr, CAP_TS6 | needcap, CAP_EBMASK, ":%s BMASK %ld %s %s :%s",
			      source_p->id, (long) chptr->channelts, chptr->chname, parv[3], degrade);
```

Move this to the right spot; where it is now will result in an out of bounds write in the `extended == false` case.

dwfreed (approved):
Don't forget to squash-merge

PR #356, opened 2022-07-27
send "End of burst" snote to L_NETWIDE

jesopo:


jillest (commented):
This is to propagate the time? Otherwise, the snote is redundant with the Netjoin one.

dwfreed (approved):

PR #357, opened 2022-08-13
OpenSSL 3.0 compatibility

jailbird777:
Attempts to clean up the deprecated functions in OpensSSL 3.0

m_challenge.c's #if's could be removed if librb/src/openssl_ratbox.h is updated
to handle EVP_MD_CTX_create/EVP_MD_CTX_destroy vs EVP_MD_CTX_new/EVP_MD_CTX_free

Only compile tested so far, as I don't have challenge setup in my environment
currently.

### DIFF
--- a/librb/src/openssl.c
+++ b/librb/src/openssl.c
@@ -392,27 +392,42 @@ rb_setup_ssl_server(const char *const certfile, const char *keyfile,
 	}
 	else
 	{
-		FILE *const dhf = fopen(dhfile, "r");
+		BIO *const dhf = BIO_new_file(dhfile, "r");
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+		EVP_PKEY* dhp = NULL;
+#else
 		DH *dhp = NULL;
+#endif
 
 		if(dhf == NULL)
 		{
-			rb_lib_log("%s: fopen ('%s'): %s", __func__, dhfile, strerror(errno));
-		}
-		else if(PEM_read_DHparams(dhf, &dhp, NULL, NULL) == NULL)
-		{
-			rb_lib_log("%s: PEM_read_DHparams ('%s'): %s", __func__, dhfile,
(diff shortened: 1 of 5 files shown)


librb/src/openssl.c
```diff
@@ -392,28 +392,28 @@ rb_setup_ssl_server(const char *const certfile, const char *keyfile,
...
 		{
-			rb_lib_log("%s: PEM_read_DHparams ('%s'): %s", __func__, dhfile,
+			rb_lib_log("%s: PEM_read_bio_Parameters ('%s'): %s", __func__, dhfile,
 			           rb_ssl_strerror(rb_ssl_last_err()));
-			fclose(dhf);
+			BIO_free(dhf);
 		}
 		else
 		{
 			SSL_CTX_set_tmp_dh(ssl_ctx_new, dhp);
```
aaronmdjones, line 411: According to [the documentation](https://www.openssl.org/docs/man3.0/man3/SSL_CTX_set_tmp_dh.html), this function is also deprecated.

It provides a suitable function to achieve the same result, `SSL_CTX_set0_tmp_dh_pkey(SSL_CTX *, EVP_PKEY *)`.

However, given that this is a `set0` function (the OpenSSL API convention for a function that assumes ownership of an object without incrementing its reference count), the object must *NOT* be freed (with `EVP_PKEY_free(3ssl)`) if this function succeeds. This function is not available in version 1.1, so it will need some `#if OPENSSL_VERSION_NUMBER` selection logic like the rest of the file.

aaronmdjones (changes requested):
I have not yet tested this PR, but see the previous comment.

librb/src/openssl.c
jailbird777: How annoying ! I think I have that cleaned up now. Still working on setting up a test server to test challenge on. Should I mark this as a draft until then?

PR #358, opened 2022-08-16
Fix SCTP support on FreeBSD & NetBSD

jailbird777:
Unlike Linux, Solaris, and Illumos (and probably others), the 2 BSDs that still support SCTP didn't put SCTP into its own library; they put it into libc.

They, unlike Linux, don't set `SOL_SCTP` for us. The official method appears to be calling `getprotobyname("sctp")` & `endprotoent()`, with `getprotobyname()` returning a struct that has a `p_proto` entry. This all reads from `/etc/protocols`. However, SCTP is assigned 132 by IANA, so it's 132 everywhere, so I just set `SOL_SCTP` to 132 if it's not already set.

### DIFF
--- a/librb/src/commio.c
+++ b/librb/src/commio.c
@@ -371,14 +371,14 @@ rb_setsockopt_sctp(rb_fde_t *F)
 	struct sctp_paddrparams paddrparams;
 	struct sctp_assocparams assocparams;
 
-	ret = setsockopt(F->fd, SOL_SCTP, SCTP_NODELAY, &opt_one, sizeof(opt_one));
+	ret = setsockopt(F->fd, IPPROTO_SCTP, SCTP_NODELAY, &opt_one, sizeof(opt_one));
 	if (ret) {
 		rb_lib_log("rb_setsockopt_sctp: Cannot set SCTP_NODELAY for fd %d: %s",
 				F->fd, strerror(rb_get_sockerr(F)));
 		return ret;
 	}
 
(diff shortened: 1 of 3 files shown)


jailbird777:
LIBSCTP_LD didn't seem to be actually used at the top-level, so I removed it there.

If you don't like the SOL_SCTP hack, I can change the #ifndef to actually do the getprotoname() lookup.

librb/src/commio.c
```diff
@@ -357,6 +357,9 @@ rb_setsockopt_reuseaddr(rb_fde_t *F)
 }
 
 #ifdef HAVE_LIBSCTP
+#ifndef SOL_SCTP
+#define SOL_SCTP 132
+#endif
 static int
```
aaronmdjones, line 363: Please insert blank lines before and after this ifndef block for enhanced readability.

aaronmdjones (changes requested):

librb/src/commio.c
jailbird777: Done, thanks!

aaronmdjones (approved):

aaronmdjones:
(Context: from discussion on IRC; various system headers (and lksctp-tools) define `IPPROTO_SCTP` to the same value)

aaronmdjones (approved):

jillest (approved):
Looks good (untested)

PR #359, opened 2022-08-16
OpenSSL 3.0 compatibility

aaronmdjones:
cf. #357

Edited by @aaronmdjones:

- Correct some data types and casts

- Minor style fixups (e.g. we put `*` on the variable name not the type)

- `librb/src/openssl.c`:

  - Defer call of `BIO_free(3ssl)` to the end of the conditional block to avoid having calls to it in multiple paths

  - Check the return value of `SSL_CTX_set0_tmp_dh_pkey(3ssl)` because if it fails then we must use `EVP_PKEY_free(3ssl)` to avoid a memory leak

    This could fail if, for example, the user supplied DSA parameters in the DH parameters file instead.

- `ircd/newconf.c`:

  - Check whether `OSSL_DECODER_CTX_new_for_pkey(3ssl)` was able to parse the given CHALLANGE public key as a valid RSA public key, and then check whether `OSSL_DECODER_from_bio(3ssl)` actually loads it successfully

- `ircd/s_newconf.c`:

  - Use `EVP_PKEY_free(3ssl)` instead of `OPENSSL_free(3ssl)` on EVP_PKEY pointers; this will avoid inadvertent memory leaks if the EVP_PKEY structure contains any dynamically-allocated child members

- `modules/m_challenge.c`:

  - Unconditionally use `EVP(3ssl)` to generate the SHA-1 digest of the random challenge; this API has been around for a very long time and is available in all supported versions of OpenSSL

  - Add lots of error checking to all steps of the process

Tested against 1.1.1 and 3.0; both with missing and provided DH parameters (which works as you'd expect; the server will not negotiate a DHE cipher without them), and CHALLENGE, including missing keys or keys of the wrong type (e.g. when you supply an EdDSA key instead of an RSA key).

This does break compatibility with OpenSSL 1.1.0 and below, which are now all end-of-life and unsupported anyway.

Closes #357

### DIFF
--- a/include/client.h
+++ b/include/client.h
@@ -215,7 +215,7 @@ struct LocalUser
 	 */
 	char *passwd;
 	char *auth_user;
-	char *challenge;
+	unsigned char *challenge;
 	char *fullcaps;
 	char *cipher_string;
 
--- a/include/s_newconf.h
+++ b/include/s_newconf.h
@@ -121,7 +121,11 @@ struct oper_conf
 
 #ifdef HAVE_LIBCRYPTO
 	char *rsa_pubkey_file;
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+	EVP_PKEY *rsa_pubkey;
+#else
 	RSA *rsa_pubkey;
+#endif
 #endif
 };
 
--- a/ircd/newconf.c
+++ b/ircd/newconf.c
@@ -4,8 +4,13 @@
 #include "stdinc.h"
 
 #ifdef HAVE_LIBCRYPTO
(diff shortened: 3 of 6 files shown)


aaronmdjones:
One thing that comes to mind (brought up on IRC) is that configuration parsing could just store the string path to the RSA public key and let `m_challenge.c` handle all of the cryptographic stuff. This would render reloading the CHALLENGE module unsafe until you restart IRCd, though. Thoughts?

aaronmdjones:
See comments on IRC about the CI commit.

dwfreed (approved):

PR #361, opened 2022-08-22
Remove ancient portability code

jailbird777:
Remove portability code for systems that don't follow at least SUSv3.
This fairly closely aligns with ISO C99, which solanum already requires.

### DIFF
--- a/librb/src/epoll.c
+++ b/librb/src/epoll.c
@@ -31,19 +31,19 @@
 #include <rb_lib.h>
 #include <commio-int.h>
 #include <event-int.h>
-#if defined(HAVE_EPOLL_CTL) && (HAVE_SYS_EPOLL_H)
+#if defined(HAVE_EPOLL_CTL) && defined(HAVE_SYS_EPOLL_H)
 #define USING_EPOLL
 #include <fcntl.h>
 #include <sys/epoll.h>
 
-#if defined(HAVE_SIGNALFD) && (HAVE_SYS_SIGNALFD_H) && (USE_TIMER_CREATE) && (HAVE_SYS_UIO_H)
+#if defined(HAVE_SIGNALFD) && defined(HAVE_SYS_SIGNALFD_H) && defined(USE_TIMER_CREATE)
 #include <signal.h>
 #include <sys/signalfd.h>
 #include <sys/uio.h>
 #define EPOLL_SCHED_EVENT 1
 #endif
 
-#if defined(USE_TIMERFD_CREATE)
+#ifdef USE_TIMERFD_CREATE
 #include <sys/timerfd.h>
 #endif
 
--- a/librb/src/linebuf.c
+++ b/librb/src/linebuf.c
@@ -545,7 +545,6 @@ rb_linebuf_flush(rb_fde_t *F, buf_head_t * bufhead)
  * autoconf checks for this..but really just want to use it if we have a
  * native version even if libircd provides a fake version...
  */
-#ifdef HAVE_WRITEV
 	if(!rb_fd_ssl(F))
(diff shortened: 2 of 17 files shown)


jailbird777:
I know large PRs are frowned upon, sorry. I can try and split out the bigger pieces (socketpair, writev, etc) if that's better.

Bumping the minimum requirement from SUSv3 -> SUSv4 would let us remove the strnlen() compatibility stuff, but I believe that's the only real savings. Personally, if I ever power up one of my Tru64 machines for nostalgia purposes, I won't shed a tear if solanum won't compile as-is on it.

Opinions?

librb/src/epoll.c
```diff
@@ -36,7 +36,7 @@
 #include <fcntl.h>
 #include <sys/epoll.h>
 
-#if defined(HAVE_SIGNALFD) && (HAVE_SYS_SIGNALFD_H) && (USE_TIMER_CREATE) && (HAVE_SYS_UIO_H)
+#if defined(HAVE_SIGNALFD) && (HAVE_SYS_SIGNALFD_H) && (USE_TIMER_CREATE)
```
aaronmdjones, line 39: Minor nitpick here (unrelated to your changes) but this should be

```C
#if defined(HAVE_SIGNALFD) && defined(HAVE_SYS_SIGNALFD_H) && defined(USE_TIMER_CREATE)
```

librb/src/sigio.c
```diff
@@ -35,7 +35,7 @@
 #include <event-int.h>
 #include <fcntl.h>		/* Yes this needs to be before the ifdef */
 
-#if defined(HAVE_SYS_POLL_H) && (HAVE_POLL) && (F_SETSIG)
+#if defined(HAVE_SYS_POLL_H) && (F_SETSIG)
```
aaronmdjones, line 38: Same here.

librb/src/epoll.c
jailbird777: Fixed!

jailbird777:
The one line in librb/configure.ac is going to merge conflict with #358 probably, but that's an easy fix.

librb/src/linebuf.c
```diff
@@ -545,13 +545,12 @@ rb_linebuf_flush(rb_fde_t *F, buf_head_t * bufhead)
...
  * native version even if libircd provides a fake version...
  */
-#ifdef HAVE_WRITEV
 	if(!rb_fd_ssl(F))
 	{
 		rb_dlink_node *ptr;
 		int x = 0, y;
 		int xret;
-		static struct rb_iovec vec[RB_UIO_MAXIOV];
+		static struct rb_iovec vec[IOV_MAX];
```
jillest, line 553: This is a step backwards in some way, since POSIX says that the `IOV_MAX` constant may not be available (in which case the application has to call `getconf(_SC_IOV_MAX)` if the guaranteed minimum `_XOPEN_IOV_MAX` is insufficient). On the other hand, not using `UIO_MAXIOV` seems good.

I think the best approach would be to call `getconf(_SC_IOV_MAX)` (once at startup) and apply some reasonable maximum such as 1024 to it. The array can still be static in this case. If `getconf(_SC_IOV_MAX)` returns -1, the value `_XOPEN_IOV_MAX` can be used. I believe using `_XOPEN_IOV_MAX` unconditionally will lead to unnecessary extra system calls, since it's only 16, it's common to send more lines than that at once and practical systems allow much more.

It may make more sense to move this part to a separate pull request.

jailbird777: Done! Thanks!

dwfreed (approved):

PR #364, opened 2022-08-23
um_callerid, um_regonlymsg: various cleanups

dwfreed:
Uses `allow_message` instead of `accept_message` so that reverse auto-`/accept` entries are only added if reverse messages would not otherwise be allowed.  This change also fixes messages to opers not triggering reverse auto-`/accept`, which is needed with the switch to a usermode (and privilege) to bypass callerid.

Also has `um_regonlymsg` check whether the source is identified before scanning the target's `/accept` list, which saves an unnecessary scan for identified sources.

### DIFF
--- a/modules/um_callerid.c
+++ b/modules/um_callerid.c
@@ -182,9 +182,7 @@ add_callerid_accept_for_source(enum message_type msgtype, struct Client *source_
 	 * as a way of griefing.  --nenolod
 	 */
 	if(msgtype != MESSAGE_TYPE_NOTICE &&
-		IsSetAnyCallerID(source_p) &&
-		!accept_message(target_p, source_p) &&
-		!IsOperGeneral(target_p))
+		!allow_message(target_p, source_p))
 	{
 		if(rb_dlink_list_length(&source_p->localClient->allow_list) <
 				(unsigned long)ConfigFileEntry.max_accept)
--- a/modules/um_regonlymsg.c
+++ b/modules/um_regonlymsg.c
@@ -85,10 +85,10 @@ allow_message(struct Client *source_p, struct Client *target_p)
 	if (IsOper(source_p))
 		return true;
 
-	if (accept_message(source_p, target_p))
+	if (source_p->user->suser[0])
 		return true;
 
-	if (source_p->user->suser[0])
+	if (accept_message(source_p, target_p))
 		return true;
 
 	return false;
--- a/modules/um_regonlymsg.c
+++ b/modules/um_regonlymsg.c
@@ -101,9 +101,7 @@ add_callerid_accept_for_source(enum message_type msgtype, struct Client *source_
 		return true;
 
 	if(msgtype != MESSAGE_TYPE_NOTICE &&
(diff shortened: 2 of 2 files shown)


aaronmdjones (dismissed):

modules/um_callerid.c
```diff
@@ -182,9 +182,7 @@ add_callerid_accept_for_source(enum message_type msgtype, struct Client *source_
 	 * as a way of griefing.  --nenolod
 	 */
 	if(msgtype != MESSAGE_TYPE_NOTICE &&
-		IsSetAnyCallerID(source_p) &&
-		!accept_message(target_p, source_p) &&
-		!IsOperGeneral(target_p))
+		!allow_message(target_p, source_p))
```
spb, line 185: I don't think this is the right test for here - if you message someone who's currently +M then an auto-accept wouldn't be added, and once +M times out they won't be able to reply. The version of this part from #383 takes account of that.

dwfreed: I think that if you have +M set, you should be around (and not just letting it time out while AFK), and equally if it gets dropped (because you're busy), you can always reset it. I also imagine the typical use of +M is such that it is set for a split second so a global/server notice can be pushed out and then cleared again. (Eg, I have an alias for OFTC's "god" mode that sets it, does the thing I want to do, and then clears it right after).  I really don't think this is going to be much of an issue in practice.

jesopo: the typical usecase is any time staff are attempting to message a user for network-related things and they're +g, in which case the only thing that needs force is your initial messages. if they choose to respond, auto-accept makes sense for a back-and-forth of unpredictable duration

dwfreed: Sure, so you set +M for your initial message, and then clear it right after, and when the user responds, auto-accept still kicks in with this change, because you are no longer +M

jesopo: why clear it when it'll time out all on its own

dwfreed: Because you don't need it anymore and you should be carrying excess privilege for as short a duration as necessary? A simple client alias can handle the whole process for you, eg `/alias overridemsg umode +M; msg $0 $1-; umode -M`

jesopo: this feels like seeking a problem to justify a solution. i dont think holding on to +M for a few minutes more than you need it is particularly a problem, same with override. real world usage would agree with me

jesopo: (especially because you can't foresee how long you'll need the powers)

dwfreed: Clogging up (and potentially completely filling) a user's /accept list with auto-accepts that they often don't even know exist isn't a problem? Especially with relaxed caller-id, every person you message, even if they are capable of messaging you because of relaxed caller-id, will earn you an auto-accept that isn't even necessary.

jesopo: no, i don't think that's a problem, or at least don't think +M is anything to do with making that not a problem

PR #375, opened 2022-10-01
Make +R play nicely with +z

spb:
This is a self-contained fix to chm_regmsg to make it work with +z, while maintaining the custom error message for -z channels.

The ideal fix might be to change the rest of the code to allow for custom errors to be sent (and suppress the standard ones in that case), but until we can work that out this change makes things work and can be (re-)loaded as a single module.


dwfreed (approved):
Approved as a quick fix; I'd like to see the `can_send` fix at some point as well, though, since `can_send` is used for more than just `PRIVMSG #channel`

PR #382, opened 2022-10-26
Note that messages caught in +g/+G are discarded

nckx:
IRC isn't e-mail (ha ha imagine oh god wait no stop) but some indication that the message will never be delivered seems welcome.

I'm in very unfamiliar territory so please let me know if this exceeds some limit, convention, or cultural norm from the 1990s.


include/messages.h
```diff
@@ -218,8 +218,8 @@
 #define NUMERIC_STR_714      ":%s 714 %s %s :You are already on that channel."
 #define NUMERIC_STR_715      ":%s 715 %s :KNOCKs are disabled."
 #define NUMERIC_STR_716      "%s :is in %s mode and must manually allow you to message them."
-#define NUMERIC_STR_717      "%s :has been informed that you messaged them."
-#define NUMERIC_STR_718      ":%s 718 %s %s %s@%s :is messaging you, and you have umode %s."
+#define NUMERIC_STR_717      "%s :has been informed that you messaged them, but your message has been discarded."
```
dwfreed, line 221: 716 would be a better numeric to modify, as that is always sent. 717 is only sent if the target is notified, which is ratelimited.

tomaw (dismissed):
+1

dwfreed (changes requested):

dwfreed (approved):

PR #388, opened 2022-11-16
chmode: ensure bants increases on banlist change

dwfreed:
This fixes an issue where a ban could be set, hit, and cleared all in the same second, and the affected client would still be cached as banned  because the bants hadn't actually changed.


jillest (commented):
Is there a need for `bants` to be a timestamp or can this just be an integer indicating the number of changes to the channel's bans?

jillest (approved):

PR #398, opened 2023-01-11
extensions/invite_notify: make the NOTICE optional, configurable

aaronmdjones:
This adds a configuration option that determines whether the NOTICE is sent to clients that do not support the IRCv3 invite-notify capability.

Requested by LiberaChat MGM.


dwfreed:
I disagree that this should be configurable to this granularity.  If invite notifications are a good and useful thing that a network wants, then imo all clients that are eligible to receive them should receive them in whatever form they are capable of understanding.  Users are free to configure their clients to handle them however they wish.

aaronmdjones:
This came up in the MGM because clients tend to react very poorly to channel notices, for example by pinging/highlighting the channel indication and creating an auditory notification.

aaronmdjones:
(There is also existing precedent for configuring how IRCd sends messages to clients to account for how they may react; `opmod_send_statusmsg`)

dwfreed:
That changes the format of the message, not whether it gets sent at all

aaronmdjones:
Yes.

Because clients can't indicate whether they will handle statusmsg properly and without going nuts, that configuration option was created so that network operators would be able to choose whether to use it or not.

Clients also can't indicate whether they'll go berserk upon receiving a channel notice. The intended behaviour here is "users in the channel will be told when someone has been invited to the channel", and not "the user's client will do everything in its power to grab the attention of the user immediately" for what is ultimately just a friendly indication and not something that needs to be acted upon.

If history has taught us anything, it's that we can't influence how clients will behave or correct problematic and/or annoying behaviour. Hell, we can't even get Thunderbird to stop blasting the user's e-mail credentials to IRC.

jillest:
Does this need to be a configuration option, or can the developers decide the right value?

Alternatively, the configuration option could be for quick experimenting/rollback (a feature flag), so it is temporary and should be removed later.

aaronmdjones:
If it were up to me, it wouldn't send a NOTICE at all. This module is named after and implements an IRCv3 capability, and thus shouldn't send anything to clients that don't support it.

As far as I understand it, LiberaChat intends to deploy this feature without the channel notice; if turning it off is not possible then I can't see us enabling this capability.

https://libera.chat/minutes/2022-11-25-public#should-we-enable-invite-notify

dwfreed (approved):

PR #406, opened 2023-04-06
support RSFNC indicating type of FNC (e.g. FORCE vs REGAIN)

jesopo:
c.f. https://github.com/atheme/atheme/commit/b2b5facd & https://github.com/atheme/atheme/pull/899

### DIFF
--- a/ircd/s_serv.c
+++ b/ircd/s_serv.c
@@ -81,6 +81,7 @@ unsigned int CAP_ENCAP;
 unsigned int CAP_TS6;
 unsigned int CAP_SERVICE;
 unsigned int CAP_RSFNC;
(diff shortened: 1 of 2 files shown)


jesopo:
for some background, atheme has two concepts of an FNC type; `FORCE` which is typically used when the target nickname wasn't chosen by the user (e.g. changing someone to `Guest`) and `REGAIN` when the target nickname was chosen by the user, usually in the case of `NS REGAIN`.

the reason the difference exists is if you `RESV` a bad nickname, you also need to go check if someone has registered that nickname and drop it if so, otherwise they can just walk over the `RESV` via `NS REGAIN`

spb (approved):

PR #410, opened 2023-06-10
Warn opers about unresponsive servers

### LINKED ISSUE #409: Snote when a remote server appears to be pinging out
Fire off a snote when a remote server has failed to provide a ping response after a certain threshold of time before it pings out. This would allow opers and any automated tooling to be aware of a laggy server that may be having issues before it splits off the network.

See also [UnrealIRCD implementation](https://github.com/unrealircd/unrealircd/blob/fd7a715e175c36f3db5271f706c493156a12b68b/src/ircd.c#L199-L209).

examknow:
Closes #409

### DIFF
--- a/doc/reference.conf
+++ b/doc/reference.conf
@@ -1315,6 +1315,10 @@ general {
 	 */
 	ping_cookie = no;
 
+	/* ping warn time: how often to warn opers of an unresponsive server
+	 */
+	ping_warn_time = 15 seconds;
+
 	/* connect timeout: sets how long we should wait for a connection
 	 * request to succeed
 	 */
--- a/doc/reference.conf
+++ b/doc/reference.conf
@@ -1315,7 +1315,10 @@ general {
 	 */
 	ping_cookie = no;
 
-	/* ping warn time: how often to warn opers of an unresponsive server
+	/* ping warn time: how long to wait after pinging a server before starting
+	 * to complain it is unresponsive. Note that the ping check interval is 30 
+	 * seconds, so the first complaint will come at the next check after this 
+	 * time has passed.
 	 */
 	ping_warn_time = 15 seconds;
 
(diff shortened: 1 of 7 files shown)


ircd/client.c
```diff
@@ -420,6 +420,24 @@ check_pings_list(rb_dlink_list * list)
 				client_p->localClient->lasttime = rb_current_time() - ping;
 				sendto_one(client_p, "PING :%s", me.name);
 			}
+			else if (!(client_p->flags & FLAGS_PINGWARN) && ConfigFileEntry.ping_warn_time > 0 &&
```
dwfreed, line 423: These warnings should repeat until the split or recovery.  This is what oftc-hybrid does, and it eliminates uncertainty about link status (and makes them much less likely to miss in the noise).  Granted, oftc-hybrid runs this every 5 seconds and solanum runs it every 30 seconds, but that's still better than nothing.

dwfreed (changes requested):

ircd/client.c
jesopo: i'd prefer it warn once when it passes the threshold and then also snote when it's back. you can infer it's still dying that way

dwfreed: Only if you actually see the snote. I have a 10 line split window for snotes; the average life of what's visible is anywhere from 15 to 60 seconds. I'd rather have an snote repeated every 30 seconds when a server's not responsive and have a better chance of seeing it than to have an snote that only fires once and then disappears off my screen in 15 seconds never to be seen again until the split happens.

doc/reference.conf
```diff
@@ -1315,6 +1315,10 @@ general {
 	 */
 	ping_cookie = no;
 
+	/* ping warn time: how often to warn opers of an unresponsive server
+	 */
```
dwfreed, line 1319: Maybe something like this?

```suggestion
	/* ping warn time: how long to wait after pinging a server before starting
	 * to complain it is unresponsive. Note that the ping check interval is 30 
	 * seconds, so the first complaint will come at the next check after this 
	 * time has passed.
	 */
```

dwfreed (approved):

PR #411, opened 2023-06-13
Add extban type for unidentified connections

TheDaemoness:
This PR adds the `extb_guest` extension. It provides the `g` extban, which acts as a normal `nick!user@host/range` ban except it only matches unidentified connections.

Channel mode `r` can be very useful for channel operators to mitigate abuse, however it is frequently too broad. Oftentimes it is only necessary to require accounts from specific ranges, allowing unregistered users from other ranges or registered users from the abused range to join unimpeded.

Currently, there does not seem to be a clean way of doing this without either `extb_combi` or auto-kicking bots/services. This extension is meant to provide a solution that can (hopefully) be loaded on Libera in the near future.


xfnw:
could you document this in the extban helptext too? (in [`help/opers/extban`](https://github.com/solanum-ircd/solanum/blob/main/help/opers/extban))

spb (approved):

PR #412, opened 2023-06-15
ircd/listener: return a fatal TLS alert for early rejected TLS clients

aaronmdjones:
This is in furtherance of commit 3fdf26aa19628d5e12a3 which added functionality to reply with a TLS record layer alert for D-Lined TLS clients. It turns out that there are other plaintext error messages in this same function that should receive the same treatment.

Also move another error string to a variable and use a compile-time optimised-out strlen for it too, to use the same approach as an existing error string.

### DIFF
--- a/ircd/listener.c
+++ b/ircd/listener.c
@@ -581,11 +581,17 @@ accept_precallback(rb_fde_t *F, struct sockaddr *addr, rb_socklen_t addrlen, voi
 	static time_t last_oper_notice = 0;
 	int len;
 
+	static const char *allinuse = "ERROR :All connections in use\r\n";
 	static const char *toofast = "ERROR :Reconnecting too fast, throttled.\r\n";
 
-	static const unsigned char sslerrcode[] = {
+	static const unsigned char ssldeniederrcode[] = {
 		// SSLv3.0 Fatal Alert: Access Denied
 		0x15, 0x03, 0x00, 0x00, 0x02, 0x02, 0x31
+	};
+
+	static const unsigned char sslinternalerrcode[] = {
+		// SSLv3.0 Fatal Alert: Internal Error
+		0x15, 0x03, 0x00, 0x00, 0x02, 0x02, 0x50
 	};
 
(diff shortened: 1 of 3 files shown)


ircd/listener.c
```diff
@@ -608,7 +609,11 @@ accept_precallback(rb_fde_t *F, struct sockaddr *addr, rb_socklen_t addrlen, voi
 			last_oper_notice = rb_current_time();
 		}
 
-		rb_write(F, "ERROR :All connections in use\r\n", 31);
+		if(listener->ssl)
+			rb_write(F, sslerrcode, sizeof(sslerrcode));
```
jillest, line 613: Alert 80 "internal_error" would be a better choice for this situation (but not any of the others below).

ircd/listener.c
```diff
@@ -656,7 +661,11 @@ accept_precallback(rb_fde_t *F, struct sockaddr *addr, rb_socklen_t addrlen, voi
```
jillest, line 661: `check_reject` above will add the client to a list; code in ircd/reject.c will later send error messages without SSL negotiation as well. Fitting in with the theme of this pull request, those should be replaced with TLS alerts as well.

ircd/reject.c
```diff
@@ -92,28 +93,41 @@ reject_exit(void *unused)
...
-
-		if (ddata->aconf)
+		if (ddata->ssl)
 		{
-			snprintf(dynamic_reason, sizeof dynamic_reason, form_str(ERR_YOUREBANNEDCREEP) "\r\n",
-				me.name, "*", get_user_ban_reason(ddata->aconf));
-			rb_write(ddata->F, dynamic_reason, strlen(dynamic_reason));
-
-			deref_conf(ddata->aconf);
+			rb_write(ddata->F, ssldeniederrcode, sizeof(ssldeniederrcode));
```
dwfreed, line 107: This misses the `deref_conf` below when `ddata->aconf` is not NULL.

dwfreed (changes requested):

dwfreed (approved):

Issue #272, opened 2021-08-24
Document extensions/filter usage


dwfreed:
There currently isn't any documentation or examples on how to get extensions/filter working.

PR #416, opened 2023-08-06
Add documentation for extensions/filter

mquin:
For #272. Adds a help file for SETFILTER and a doc/features article explaining how the filter extension works.


glguy:
Is it obvious if the chunking happens before or after base64 encoding? If not could we elaborate on that?

glguy (approved):
The content looks correct and it's great to get this documented!

glguy (approved):

PR #417, opened 2023-08-23
document +M cmode

Svetlana-T:
- as per mention of it in #libera by moonmoon 
- maybe there are other additional cmodes not documented, didn't check

### DIFF
--- a/help/opers/cmode
+++ b/help/opers/cmode
@@ -37,7 +37,7 @@ NO PARAMETERS:
      +C     - Disable CTCP. All CTCP messages to the channel, except ACTION,
               are disallowed.
   ?  +O     - IRC Operator only channel.
-     +M     - IRC Operators can not be kicked. Only settable by opers. Only
+  ?  +M     - IRC Operators can not be kicked. Only settable by opers. Only
               viewable by opers.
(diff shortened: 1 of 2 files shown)


help/opers/cmode
```diff
@@ -2,6 +2,8 @@ MODE <channel> <+|-><modes> [parameters]
 
 ? designates that the cmode is provided by an extension
 and may not be present on this server.
+
+* designates that the cmode is oper only.
```
glguy, line 6: We don't need a new symbol for this. It's true for other modes too, like `L` and `P` and the goes into the comment about it.

Svetlana-T: @glguy - Fixed below - let me know if any further corrections are suggested

help/opers/cmode
```diff
@@ -37,6 +37,8 @@ NO PARAMETERS:
      +C     - Disable CTCP. All CTCP messages to the channel, except ACTION,
               are disallowed.
   ?  +O     - IRC Operator only channel.
+     +M     - IRC Operators can not be kicked. Only settable by opers. Only
```
glguy, line 40:
```suggestion
  ?  +M     - IRC Operators can not be kicked. Only settable by opers. Only
```

PR #418, opened 2023-09-03
client: refactor del_all_accepts to allow skipping own accept list

dwfreed:
This allows reusing this function for other uses that just need to remove this client from others' accept lists on nick change and not have duplicates of this code everywhere


glguy (approved):
Logic makes sense to me

PR #420, opened 2023-10-04
Improve error messages on OPER/CHALLENGE failure

aaronmdjones:


jillest (commented):
These may not have been omissions but deliberately ensuring existing monitor bots and scripts kept working. EFNet had quite a few of them.

This doesn't mean solanum should not improve this, though.

spb (approved):

PR #421, opened 2023-11-05
Delay rehashing until we're not processing events

edk0:
Fixes bug/probable regression introduced by 0ab6dbbc651ddd1c26cb7baa6e6cf86890a4abd2

rehash() closes listeners. If we happen to get a single epoll() result that wants to first rehash and then accept a connection, the epoll info will point to a freed rb_fde_t. Other selectors should have similar problems, but we didn't investigate that.

rb_fde_ts are normally batched up and freed outside the event processing, but as of the above commit close_listeners() screws that up by closing pending FDs immediately in order to create new listeners.

I think it might be a bit better to revert this behaviour and simply not close listeners if we are going to open new ones over them, but have opted for the smallest reasonable change I can think of.

### DIFF
--- a/ircd/s_conf.c
+++ b/ircd/s_conf.c
@@ -642,16 +642,18 @@ attach_conf(struct Client *client_p, struct ConfItem *aconf)
 	return (0);
 }
 
-/*
- * rehash
- *
- * Actual REHASH service routine. Called with sig == 0 if it has been called
- * as a result of an operator issuing this command, else assume it has been
- * called as a result of the server receiving a HUP signal.
- */
-bool
-rehash(bool sig)
+struct rehash_data {
+	bool sig;
+};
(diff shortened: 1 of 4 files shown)


ircd/s_conf.c
```diff
@@ -642,16 +642,18 @@ attach_conf(struct Client *client_p, struct ConfItem *aconf)
...
+	bool sig;
+};
+
+static void
+service_rehash(void *data_)
 {
+	struct rehash_data *data = data_;
+	bool sig = data->sig;
+
+	free(data);
```
aaronmdjones, line 655: This should be `rb_free`, no?

glguy:
```suggestion
	rb_free(data);
```

PR #424, opened 2023-11-27
Add function parameter prototypes for clang

jailbird777:
Modern clang (16+) requires all function prototypes to have arguments.

This affects the old K&R style in librb/src/crypt.c and the 'DCF' declaration

### DIFF
--- a/authd/authd.c
+++ b/authd/authd.c
@@ -199,7 +199,7 @@ main(int argc, char *argv[])
 	rb_set_time();
 	setup_signals();
 
-	authd_option_handlers = rb_dictionary_create("authd options handlers", rb_strcasecmp);
+	authd_option_handlers = rb_dictionary_create("authd options handlers", (DCF)rb_strcasecmp);
 
 	init_resolver();
 	init_providers();
--- a/tests/rb_dictionary1.c
+++ b/tests/rb_dictionary1.c
@@ -33,7 +33,7 @@
 
 static void replace1(void)
 {
-	rb_dictionary *dict = rb_dictionary_create("replace1", strcmp);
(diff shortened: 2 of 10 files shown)


authd/authd.c
```diff
@@ -199,7 +199,7 @@ main(int argc, char *argv[])
 	rb_set_time();
 	setup_signals();
 
-	authd_option_handlers = rb_dictionary_create("authd options handlers", rb_strcasecmp);
+	authd_option_handlers = rb_dictionary_create("authd options handlers", (DCF)rb_strcasecmp);
```
jillest, line 202: Casting function pointers depends on the representation of `const char *` and `const void *` being the same, which is likely but not guaranteed. It would be safer to have wrappers (one for `rb_strcasecmp` and one for `irccmp`) that can be passed to `rb_dictionary_create` without a conversion.

Since `rb_strcasecmp` is part of librb, librb could also provide this wrapper. The wrapper for `irccmp` will need to be in ircd/.

By the way, the unsafety is similar to the previous unsafety since the code in librb/src/dictionary.c calls the unprototyped function with two `const void *` in all cases. The change to a prototyped function adds checking to those calls.

tests/rb_dictionary1.c
```diff
@@ -33,7 +33,7 @@
 
 static void replace1(void)
 {
-	rb_dictionary *dict = rb_dictionary_create("replace1", strcmp);
+	rb_dictionary *dict = rb_dictionary_create("replace1", (DCF)strcmp);
```
jillest, line 36: This `strcmp` also needs a wrapper somewhere to fix the types.

jillest (changes requested):

aaronmdjones:
Obsoleted by recent changes to the CI configuration necessitating the same.

PR #427, opened 2024-03-16
Remove WebSocket support

aaronmdjones:
Discussion on removing this has come up numerous times in the past, with the general consensus being that it is not wanted and does not work properly, but nothing has actually happened as far as trying to remove it before. Should we still decide we want to go ahead and remove it, this PR accomplishes that.


dwfreed (approved):
The original thought was to hold off on removing it until we could offer a replacement.  But Libera uses the kiwi websocket server, and OFTC will have a websocket server figured out as part of upgrading to solanum, so it's not really an issue.

spb (approved):

tomaw (approved):

ilbelkyr (approved):

PR #430, opened 2024-06-11
Fix WHOIS reply not respecting a lack of the multi-prefix capability

JustAnotherArchivist:
Prior to this commit, `RPL_WHOISCHANNELS` always contained all channel membership prefixes, even when the client had not enabled the multi-prefix capability. This behaviour predates available commit history. Other ircds only return the highest prefix, as prescribed by the RFCs.

(Thanks to @fireonlive for helping me test this!)

### DIFF
--- a/modules/m_whois.c
+++ b/modules/m_whois.c
@@ -279,7 +279,7 @@ single_whois(struct Client *source_p, struct Client *target_p, int operspy)
 			{
 				send_multiline_item(source_p, "%s%s%s",
 						hdata_vis.approved ? "" : "!",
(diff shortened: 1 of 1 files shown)


aaronmdjones (approved):

aaronmdjones:
Unfortunately I have had to revert this due to reasons explained in commit 746ced2308f606ddaa2781ada5eac4e4364b8f3f

JustAnotherArchivist:
Ah, that's unfortunate. I'll open an issue for tracking this instead.

PR #432, opened 2024-06-12
m_whois: restore original multi-prefix behavior

dwfreed:
The original behavior has existed since before git history, and has no known impact on clients that exist today.

This reverts commit 746ced2


glguy (approved):
Yes, please.

Issue #334, opened 2022-05-09
pre-sanitise usernames are logged in SNO_BANNED


jesopo:
assigning to edk because I think he knew exactly how this worked

PR #433, opened 2024-06-13
Filter [ out of Rejecting K-Lined user snote

glguy:
This snote is generated earlier than any other snote that called get_client_name which causes it to potentially have the wrong username in the message. The extra [ can result in an ambiguous parse as well.


glguy:
Related to #334

PR #434, opened 2024-06-13
Sanitize username early, and early reject invalid ones

### LINKED ISSUE #334: pre-sanitise usernames are logged in SNO_BANNED

dwfreed:
This ensures [ won't appear in the various rejection snotes, and invalid username characters only appear in the snote announcing their rejection.

Supersedes #433 and #381
Fixes #334

### DIFF
--- a/ircd/s_conf.c
+++ b/ircd/s_conf.c
@@ -82,8 +82,8 @@ static int cmp_prop_ban(const void *, const void *);
 FILE *conf_fbfile_in;
 extern char yytext[];
 
-static int verify_access(struct Client *client_p, const char *username);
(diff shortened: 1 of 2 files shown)


glguy:
I believe that we need to test that the new `[`-free username is not empty now. Before the change I get "invalid username" trying this. After the change I seem to be able to get an empty username

dwfreed:
Fixed by moving the `valid_username` test to after the sanitization.  It'll look like they have an empty username in the snote, but that's no worse than it was before.

glguy (approved):
Works in my test instance; format strings seem correct

tomaw (approved):

PR #443, opened 2025-01-29
Don't set the oper flags on client

glguy:
The oper_conf flags are a different set of flag
bits than the client flags. The oper flags are
hardcoded to have ENCRYPTED set which is the same
flag as PINGSENT on the client struct. This can
cause a client to timeout if the last thing it
does is oper up.

### DIFF
--- a/include/s_newconf.h
+++ b/include/s_newconf.h
@@ -140,8 +140,6 @@ extern void cluster_generic(struct Client *, const char *, int cltype,
 #define OPER_ENCRYPTED	0x00001
 #define OPER_NEEDSSL    0x80000
 
-#define OPER_FLAGS	0	 /* no oper privs in Client.flags/oper_conf.flags currently */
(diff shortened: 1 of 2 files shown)


ilbelkyr:
It looks like this was a remnant of a mechanism to apply privileges directly from the operator blocks which was deprecated with the introduction of privsets and removed in 10847f65d. The deoper code still makes sure to remove oper-only flags:

https://github.com/solanum-ircd/solanum/blob/4e89f6603d63b2f8dfdec7a83161fb0343236349/include/s_newconf.h#L143
https://github.com/solanum-ircd/solanum/blob/4e89f6603d63b2f8dfdec7a83161fb0343236349/ircd/s_user.c#L1180

We can probably remove those lines as well considering they've gone unused since 2008.

ilbelkyr (approved):

PR #444, opened 2025-02-11
Add max_autoconn to documentation

badeiomme:
7c7065b0 introduced class::max_autoconn configuration parameter. However, this parameter was never added to config.rst and the example config. This commits adds them there as well.


aaronmdjones:
`max_number` should not have been removed in this. Its description is incorrect, but it is still checked. It should instead say that it is the maximum number of servers that can be connected, rather than how many will be auto connected.

PR #446, opened 2025-02-14
GitHub Docs Workflow: Bump runner version

aaronmdjones:


.github/workflows/docs.yaml
```diff
@@ -11,11 +11,10 @@ on:
       - main
     paths:
       - 'doc/oper-guide/**'
-      
+
 jobs:
   build:
-    runs-on: ubuntu-18.04
-
+    runs-on: ubuntu-24.04
```
dwfreed, line 17:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/docs.yaml
```diff
@@ -24,6 +23,9 @@ jobs:
             python-sphinx \
             texinfo \
             # EOF
-      - uses: actions/checkout@v2
-      - name: Build
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          submodules: recursive
```
dwfreed, line 29:
```suggestion
      - uses: actions/checkout
```

The action settings don't matter for us, no real point in specifying them.

dwfreed (changes requested):
I know one normally wouldn't use latest and non-fixed action versions, but we update this so rarely that it's probably easier to just let it break down the road and fix it when that happens than have to remember to bump these periodically.

.github/workflows/docs.yaml
aaronmdjones: I am not comfortable removing `persist-credentials`. This is a recipe for being bitten in the future with how a modern `actions/checkout` behaves, for example if we want to make this workflow one day publish the documentation somewhere.

https://github.com/actions/checkout/issues/485

dwfreed (approved):

PR #450, opened 2025-05-02
Add dedicated hook data type for can_create_channel

TheDaemoness:
`hook_data_can_create_channel` includes the name of the channel being created, which is necessary for any non-trivial restriction of channel creation as well as sending correct error messages.

The new hook data matches the layout of the previous hook data, taking advantage of a disused pointer field in `hook_data_client_approval` as used by callers of the hook (`m_join`).

There are two in-tree extensions that use can_create_channel. This PR doesn't currently attempt to change either of them to benefit from the new functionality.

### DIFF
--- a/extensions/createauthonly.c
+++ b/extensions/createauthonly.c
@@ -31,7 +31,7 @@ DECLARE_MODULE_AV2(createauthonly, NULL, NULL, NULL, NULL, restrict_hfnlist, NUL
 static void
 h_can_create_channel_authenticated(void *data_)
 {
-	hook_data_client_approval *data = data_;
+	hook_data_can_create_channel *data = data_;
 	struct Client *source_p = data->client;
 
 	if (*source_p->user->suser == '\0' && !IsOperGeneral(source_p))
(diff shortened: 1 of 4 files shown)


dwfreed:
@jillest Do you recall why this wasn't done previously? The hook call predates git history

jillest:
> `hook_data_can_create_channel` includes the name of the channel being created, which is necessary for any non-trivial restriction of channel creation as well as sending correct error messages.

Where "non-trivial" means "depending on the channel name".

> The new hook data matches the layout of the previous hook data, taking advantage of a disused pointer field in `hook_data_client_approval` as used by callers of the hook (`m_join`).

OK.

> There are two in-tree extensions that use can_create_channel. This PR doesn't currently attempt to change either of them to benefit from the new functionality.

Why would there need to be a change to createauthonly and createoperonly? They work fine as they are. As I see it, the new functionality would be used in a new module.

> @jillest Do you recall why this wasn't done previously? The hook call predates git history

My guess is that only use cases like createauthonly and createoperonly were being thought of, even though `##` channels might have already existed on freenode back then (before 2007). Formal channel namespaces did not exist, and blocking creation of any `#` (not `##`) channel would have been too heavy-handed.

jillest (approved):

TheDaemoness:
> Why would there need to be a change to createauthonly and createoperonly? They work fine as they are.

Except that the message in the `ERR_NEEDREGGEDNICK` (477) reply template implies that cmode +r is the cause of the failure to join, which is misleading when used here. If you consider that "fine", well… agree to disagree.

Either way, I appreciate the approval and merge. Thank you!

PR #456, opened 2025-07-16
Add bypass capability to spamfilter

skizzerz:
This adds a new flag that causes other actions attached to that filter to apply even if the target user or channel is +u. In other words, it is not possible to "opt out" of filters using the bypass flag. Opers remain immune, both as senders and recipients, so that spamfilter cannot be abused to lock people out of managing filters, talking to services, etc.

By default, this flag only works if combined with the "drop" action however a new ircd.conf setting can be used to allow it to apply to "kill" and "alarm" as well. Another new ircd.conf setting can expose the nick!user@host of the sender to the filtering engine (previously this depended on changing compile-time macros). If either of these settings are enabled, a new user-visible "F" flag will appear in VERSION output to indicate that the spamfilter engine has these capabilities.


aaronmdjones (dismissed):

aaronmdjones:
@skizzerz  The code changes in this pull request do not match the table you just provided for the ALARM case.

If it should be as you described you are missing 2 instances of

```diff
-if (r & ACT_ALARM) {
+if (r & ACT_ALARM && mask != ACT_SUPERDROP) {
```

dwfreed (changes requested):
As actions are not mutually exclusive, I would rather this operated as a flag that meant "Ignore +u when applying actions" so that alerts and kills also still happened if appropriate for the given pattern.

skizzerz:
@dwfreed That doesn't really match the desired goal here, which is that +u can only be overridden via the drop action, and kill/alarm will always respect +u preference. Drop has no oper-visible tells, unlike kill/alarm, and as such cannot be abused as a dysfunctional spying platform. While I wasn't around at the time it was introduced, I heard that concerns about potential oper abuse of the filters is (partly) why +u exists in the first place, and letting opers render the mode obsolete would remove that privacy protection.

bakerst-221b:
In my opinion there should be a firm restriction in the IRCD that prevents the "alarm" action (and to a lesser degree also the "kill" action) from applying to +u targets.

The reason why I think this is a requirement is to show good faith by not implementing abuse-prone features. The IRCD is open source, while the tools used for creating and uploading databases aren't. Having the restriction in the IRCD allows networks to point at the source code and say: "We specifically designed this feature so that the IRCD doesn't allow opers to bypass +u and spy on your private messages using this feature."

The superdrop flag / this PR achieves that.

I suppose we could also use a flag to implement this but rather than saying "Ignore +u when applying actions", that flag would need to say "Ignore +u when applying the drop action".

jillest:
All this complexity seems only useful if there is a sufficiently large network that will use `filter_sees_user_info = off` and `filter_bypass_all = off`. Otherwise, we will just meaninglessly have this `F` everywhere.

aaronmdjones (approved):

PR #465, opened 2025-10-20
Fix memory leak in sendto_match_servs

skizzerz:
Free the msgbuf cache after we're done.


andir (approved):
Been running this on hackint for >12h now. The memory usage this morning is identical with the one from ~8h ago when I headed to bed. Seems like it fixed it for us. Thanks! <3

PR #466, opened 2025-11-05
Add batch support

skizzerz:
This has been tested alongside an implementation of draft/multiline (not part of this PR due to being a draft) to verify that netsplit/netjoin batches work as intended as well as the m_batch module for c2s and s2s batches.

- Add "batch" client capability
- Support sending netsplit/netjoin batches
- Support receiving BATCH from clients and remote servers
- Support for modules to define supported batch types for incoming batches
- Add "message_handler" hook to override the handler for a particular incoming message
- Add technical and user-facing documentation for batches
- Add some more sendto_* overloads that support tags which were missing from the initial message-tags implementation but turned out to be needed

### DIFF
--- a/modules/m_batch.c
+++ b/modules/m_batch.c
@@ -81,7 +81,7 @@ batch_timeout(void *arg)
 			{
 				sendto_one(client, ":%s FAIL BATCH TIMEOUT %s :Batch timed out",
 					me.name, batch->tag);
-				client->localClient->pending_batch_lines -= batch->len;
+				client->localClient->pending_batch_lines -= batch->len + 1;
 				batch_free(batch);
 				rb_dlinkDestroy(ptr, &client->localClient->pending_batches);
 			}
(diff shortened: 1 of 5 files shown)


modules/m_batch.c
```diff
@@ -0,0 +1,388 @@
...
+			return;
+		}
+	}
+}
+
+static void
+m_batch(struct MsgBuf *msgbuf, struct Client *client_p, struct Client *source_p, int parc, const char *parv[])
+{
+	struct Batch *batch = NULL, *parent = NULL;
+	rb_dlink_node *ptr, *next_ptr;
```
aaronmdjones: This variable is not used

```
m_batch.c:222:23: warning: unused variable 'next_ptr' [-Wunused-variable]
  222 |         rb_dlink_node *ptr, *next_ptr;
      |                              ^~~~~~~~
```

Did you mean to use the `_SAFE` variant below on lines 230 & 266? I doubt it; neither loop body modifies the list they're operating on.

skizzerz: This was probably a holdover from a previous design; I went through like 4 iterations internally of how all of this was structured before settling on this one and making the PR. I'll remove the unused variable.

skizzerz:
Tried this out on my testnet and fixed issues I've encountered with it. I think this is ready for proper review now.

Proper comparison link since some intervening commits had some testing garbage in them: https://github.com/solanum-ircd/solanum/compare/34231f59590c1c15ea229f6e6d3ae18bbd65d4ae..47acb185f34409dd5201524619dfb9ad85561a4d

include/client.h
```diff
@@ -275,6 +275,9 @@ struct LocalUser
 	unsigned int sasl_messages;
 	unsigned int sasl_failures;
 	time_t sasl_next_retry;
+
+	uint32_t pending_batch_lines;	/* number of lines held in pending client-initiated batches */
+	rb_dlink_list pending_batches;	/* batches this client started but did not finish */
```
A-UNDERSCORE-D: Should this rather be something like `alive batches that this client...`

aaronmdjones (approved):
Tentative approval; I can't find anything wrong with it and neither ASan nor Valgrind complains with message batches.

PR #467, opened 2025-11-15
umode +M (override +g) should snotice on use

### LINKED ISSUE #451: umode +M (override +g) should snotice on use
Given you're explicitly choosing to ignore a user's decision to have callerid, we should not allow this to happen silently

rpiadam:
Fixes #451: When an oper uses umode +M to override callerid (+g/+G), send a server notice to all opers so the action is logged and visible.

- Added snotice when +M is set in check_umode_change()
- Added snotice when +M is actually used to override callerid in allow_message()

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -34,18 +34,26 @@ include/serno.h:
 		echo '#define DATECODE 0UL' >>include/serno.h; \
 	fi
 
-install-data-hook:
+install-data-hook: install-data
 	test -d ${DESTDIR}${logdir} || mkdir -p ${DESTDIR}${logdir}
 
-install-exec-hook:
-	rm -f ${DESTDIR}${libdir}/*.la
-	rm -f ${DESTDIR}${moduledir}/*.la
-	rm -f ${DESTDIR}${moduledir}/autoload/*.la
-	rm -f ${DESTDIR}${moduledir}/extensions/*.la
-	rm -f ${DESTDIR}${libdir}/*.dll.a
-	rm -f ${DESTDIR}${moduledir}/*.dll.a
-	rm -f ${DESTDIR}${moduledir}/autoload/*.dll.a
-	rm -f ${DESTDIR}${moduledir}/extensions/*.dll.a
+install-exec-hook: install-exec
+	@if test -d ${DESTDIR}${libdir}; then \
+		rm -f ${DESTDIR}${libdir}/*.la; \
+		rm -f ${DESTDIR}${libdir}/*.dll.a; \
+	fi
+	@if test -d ${DESTDIR}${moduledir}; then \
+		rm -f ${DESTDIR}${moduledir}/*.la; \
+		rm -f ${DESTDIR}${moduledir}/*.dll.a; \
+	fi
+	@if test -d ${DESTDIR}${moduledir}/autoload; then \
+		rm -f ${DESTDIR}${moduledir}/autoload/*.la; \
+		rm -f ${DESTDIR}${moduledir}/autoload/*.dll.a; \
+	fi
+	@if test -d ${DESTDIR}${moduledir}/extensions; then \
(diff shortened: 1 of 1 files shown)


modules/core/m_message.c
```diff
@@ -832,7 +832,10 @@ echo_msg(struct Client *source_p, struct Client *target_p,
 
 	if (MyClient(target_p))
 	{
-		if (!IsCapable(target_p, CLICAP_ECHO_MESSAGE))
+		/* Always echo messages to services, regardless of capability */
+		bool is_service = IsService(target_p);
+		
+		if (!IsCapable(target_p, CLICAP_ECHO_MESSAGE) && !is_service)
```
jillest, line 870: Please do not add unrelated fixes. This is unrelated to overriding +g.

modules/um_callerid.c
```diff
@@ -248,8 +249,32 @@ h_hdl_privmsg_user(void *vdata)
...
+	if (MyClient(target_p) && IsSetAnyCallerID(target_p) && IsPerson(source_p))
+	{
+		if (!(IsSetRelaxedCallerID(target_p) &&
+				!IsSetStrictCallerID(target_p) &&
+				has_common_channel(source_p, target_p)) &&
+			!accept_message(source_p, target_p) &&
+			!MayHavePrivilege(source_p, "oper:always_message"))
+		{
+			would_block = true;
+		}
```
jillest, line 263: This duplicates the logic in `allow_message`, which is both inefficient and makes it more likely that future changes will not be applied correctly to both copies. Please move this determination to `allow_message`, making it return one out of three possibilities.

modules/um_callerid.c
```diff
@@ -248,8 +249,32 @@ h_hdl_privmsg_user(void *vdata)
...
+
 	if (allow_message(source_p, target_p))
+	{
+		/* If +M was the reason the message was allowed, send snotice */
+		if (would_block && IsSetTalkThroughCallerID(source_p) && MyClient(source_p))
+		{
+			sendto_realops_snomask(SNO_GENERAL, L_ALL,
+					"%s (%s@%s) used umode +M to override %s's +%s",
+					source_p->name, source_p->username, source_p->host,
+					target_p->name, IsSetStrictCallerID(target_p) ? "g" : "G");
```
jillest, line 274: This might become noisy if the conversation becomes longer. I don't immediately know how to fix it, though.

Makefile.am
```diff
@@ -34,18 +34,26 @@ include/serno.h:
 		echo '#define DATECODE 0UL' >>include/serno.h; \
 	fi
 
-install-data-hook:
+install-data-hook: install-data
```
jillest, line 37: OK, but it belongs in a separate branch and pull request.

Makefile.am
```diff
@@ -34,18 +34,26 @@ include/serno.h:
...
+		rm -f ${DESTDIR}${moduledir}/*.dll.a; \
+	fi
+	@if test -d ${DESTDIR}${moduledir}/autoload; then \
+		rm -f ${DESTDIR}${moduledir}/autoload/*.la; \
+		rm -f ${DESTDIR}${moduledir}/autoload/*.dll.a; \
+	fi
+	@if test -d ${DESTDIR}${moduledir}/extensions; then \
+		rm -f ${DESTDIR}${moduledir}/extensions/*.la; \
+		rm -f ${DESTDIR}${moduledir}/extensions/*.dll.a; \
+	fi
```
jillest, line 56: I have no idea why this change is useful since `rm -f` already ignores errors from nonexistent files.

rpiadam:
ya all was fixed one go and first was was allready pushed a pull request that

modules/core/m_message.c
rpiadam: I made u system more secure then you ever have 📦

PR #468, opened 2025-11-15
Fix/426 316 extb canjoin improvements

rpiadam:


extensions/extb_canjoin.c
```diff
@@ -63,7 +64,51 @@ static int eb_canjoin(const char *data, struct Client *client_p,
...
+		/* If the ban has a forward, check if user would be banned in the forward channel */
+		if (forward != NULL)
+		{
+			struct Channel *forward_chptr = find_channel(forward);
+			if (forward_chptr != NULL && forward_chptr != chptr && forward_chptr != chptr2)
+			{
+				/* Check if user is banned in forward channel (but don't recurse $j checks) */
+				recurse = 0;
+				int forward_banned = is_banned(forward_chptr, client_p, NULL, &ms, NULL) != 0 ||
+				                     is_quieted(forward_chptr, client_p, NULL, &ms) == CHFL_BAN;
```
ilbelkyr, line 83: for what it's worth, it would seem inconsistent to take ban exceptions into account for quiets when we deliberately aren't doing so for bans

extensions/extb_canjoin.c
```diff
@@ -63,7 +64,51 @@ static int eb_canjoin(const char *data, struct Client *client_p,
...
+	if (banned_result != 0)
+	{
+		/* If the ban has a forward, check if user would be banned in the forward channel */
+		if (forward != NULL)
+		{
+			struct Channel *forward_chptr = find_channel(forward);
+			if (forward_chptr != NULL && forward_chptr != chptr && forward_chptr != chptr2)
+			{
+				/* Check if user is banned in forward channel (but don't recurse $j checks) */
+				recurse = 0;
```
ilbelkyr, line 81: `recurse` is used to track whether we are checking bans from inside a `$j`. Unsetting it here will cause the opposite of what's intended and allow for (potentially infinite) recursion.

extensions/extb_canjoin.c
```diff
@@ -63,7 +64,51 @@ static int eb_canjoin(const char *data, struct Client *client_p,
...
+		{
+			ret = EXTBAN_MATCH;
+		}
+	}
+	else
+	{
+		/* Check quiets separately to also check for forward */
+		const char *qforward = NULL;
+		int quieted_result = is_quieted(chptr2, client_p, NULL, &ms);
+		if (quieted_result == CHFL_BAN)
```
ilbelkyr, line 102: `qforward` is unused.

The above comment about applying ban exceptions applies here as well.

PR #469, opened 2025-11-15
Fix/362 notify q on forward

rpiadam:


modules/core/m_join.c
```diff
@@ -99,10 +99,33 @@ check_forward(struct Client *source_p, struct Channel *chptr,
...
+	/* User is +Q - check if a forward would have been attempted and notify */
+	if (IsNoForward(source_p))
+	{
+		const char *next_test = NULL;
+		struct Channel *test_chptr;
+		int test_err;
+		
+		/* Check if there would have been a forward available */
+		test_err = can_join(source_p, chptr, key, &next_test);
+		if (test_err != 0 && next_test != NULL)
```
ilbelkyr, line 115: We only get to this point if `can_join` with equivalent arguments returned nonzero. `next_test` is also redundant as it will match `next` as populated by the original `can_join` call.

modules/core/m_join.c
```diff
@@ -99,10 +99,33 @@ check_forward(struct Client *source_p, struct Channel *chptr,
...
+		{
+			test_chptr = find_channel(next_test);
+			if (test_chptr != NULL && !IsMember(source_p, test_chptr) &&
```
ilbelkyr, line 118: I'm unsure this should check for channel membership, tbh – it still seems sensible to notify users about the forward even in cases where they're already on the target channel. We already have a special case for forwards that fail due to being on the target channel (using `ERR_USERONCHANNEL` instead of the reason joining the original channel would otherwise fail) and I can't think of a good reason we shouldn't notify `+Q` users under the same circumstances.

modules/core/m_join.c
```diff
@@ -99,10 +99,33 @@ check_forward(struct Client *source_p, struct Channel *chptr,
...
+			    !hash_find_resv(test_chptr->chname) &&
+			    !(test_chptr->mode.mode & MODE_DISFORWARD))
+			{
+				/* A forward would have been attempted, notify the user */
+				sendto_one_notice(source_p, ":*** Forward to %s was blocked due to +Q mode", test_chptr->chname);
```
ilbelkyr, line 123: This notice should probably mention which channel the user originally tried to join.

(A numeric would probably be even better, but I don't know if there's any existing numeric – possibly one we could steal from another ircd – that'd fit?)

modules/core/m_join.c
```diff
@@ -99,10 +99,33 @@ check_forward(struct Client *source_p, struct Channel *chptr,
...
+		/* Check if there would have been a forward available */
+		test_err = can_join(source_p, chptr, key, &next_test);
+		if (test_err != 0 && next_test != NULL)
+		{
+			test_chptr = find_channel(next_test);
+			if (test_chptr != NULL && !IsMember(source_p, test_chptr) &&
+			    !hash_find_resv(test_chptr->chname) &&
+			    !(test_chptr->mode.mode & MODE_DISFORWARD))
+			{
+				/* A forward would have been attempted, notify the user */
```
ilbelkyr, line 122: The comment above this function explicitly states that it does not send any notices to the user. IMHO it's probably more elegant to stick to this and adjust the signature to let it communicate this circumstance to the caller?

PR #479, opened 2025-12-07
Replace git log usage with git rev-list

hello-smile6:
As explained in libera #solanum:

```text
<h|weechat> Just discovered that Makefile.am:26 " datecode=`git log -1
--pretty=format:%ct`; \" (which gets fed into serno.h) silently
malfunctions and includes git signature-related data in the generated
serno.h in some conditions, including when the latest commit is signed,
if the git config option log.showsignature is set to true, breaking the
build process of solanum. Is a PR that fixes this welcome?
```

This change should not cause any regressions on old versions of git, as,
when tested on git 2.44.1, git appears to ignore invalid config
options.

### DIFF
--- a/scripts/makerelease.sh
+++ b/scripts/makerelease.sh
@@ -26,7 +26,7 @@ else
 	RELEASENAME="$1"
 fi
 
-TIP=`git -c log.showsignature=false log -1 --pretty=oneline | cut -d" " -f1`
+TIP=`git rev-list -1 --no-commit-header --format=%h HEAD`
 
 WRKDIR=`pwd`
 


scripts/makerelease.sh
```diff
@@ -26,7 +26,7 @@ else
 	RELEASENAME="$1"
 fi
 
-TIP=`git log -1 --pretty=oneline | cut -d" " -f1`
+TIP=`git -c log.showsignature=false log -1 --pretty=oneline | cut -d" " -f1`
```
dwfreed:
```suggestion
TIP=`git rev-list -1 --no-commit-header --format=%h HEAD`
```

dwfreed (changes requested):

PR #487, opened 2026-01-22
Add doing_version_confopts hook

skizzerz:
This hook allows modification of the confopts portion of a VERSION response (i.e. the set of flags representing how the ircd was configured).

Extension modules can use this hook to display additional confopts without needing to edit m_version or add things to core in order to share state with m_version.

### DIFF
--- a/modules/m_version.c
+++ b/modules/m_version.c
@@ -38,6 +38,7 @@ static const char version_desc[] =
 	"Provides the VERSION command to display server version information";
 
 static char *confopts(void);
+static int h_doing_version_confopts;
 
 static void m_version(struct MsgBuf *, struct Client *, struct Client *, int, const char **);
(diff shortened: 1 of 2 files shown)


modules/m_version.c
```diff
@@ -113,34 +119,49 @@ mo_version(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *sour
...
 confopts(void)
 {
-	static char result[15];
-	char *p;
-
-	result[0] = '\0';
-	p = result;
+	static char result[64];
+	char *p = result;
+	char c;
```
aaronmdjones:
```C
char c;
```

This should be int, not char. `typeof 'a' == int` and `tolower(3)` and the like take an `int`.

aaronmdjones (approved):

PR #493, opened 2026-02-02
Add m_quarantine

skizzerz:
This extension module provides a new usermode (+q) and two commands that allow opers to set/unset that mode. The mode can be set either via ircd.conf (general::default_umodes or auth::umodes) or via the QUARANTINE command. When using the command, a reason is required which will be broadcast to all other opers via the snote.

When set, a quarantined user will not be able to join channels or send messages to channels, with exception of explicitly allowlisted channels in ircd.conf. They will also not be able to send private messages to other users except opers or services.

The quarantine will be automatically removed if the target opers up or logs into a services account, and opers and logged in users cannot be quarantined. The UNQUARNTINE command may additionally be used to remove the quarantine status from a user. Users cannot otherwise set or remove this usermode from themselves.

The QUARANTINE and UNQUARANTINE commands require a new priv oper:quarantine.

To support this, a new account_change hook was added to m_services and m_signon. This hook is fired whenever a user's services account changes (via logging in or logging out) from the SVSLOGIN, SIGNON, LOGIN, or SU services commands. It is **not** fired for SVSLOGIN on unknown (pre-registration) users.

Documentation was updated for the new usermode, hook, and commands. Usermode help was alphabetized for consistency, a missing entry for helpops was added, and cloaking was moved to +x (since all modern cloaking modules use that mode instead) since I was touching the file anyway.

### DIFF
--- a/extensions/m_quarantine.c
+++ b/extensions/m_quarantine.c
@@ -1,0 +1,570 @@
+/*
+ * Solanum: a slightly advanced ircd
+ * m_quarantine.c: restrict connections until they identify to services
+ *
+ * Copyright (c) 2026 Ryan Schmidt <skizzerz@skizzerz.net>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "stdinc.h"
+#include "channel.h"
+#include "chmode.h"
+#include "modules.h"
+#include "hook.h"
+#include "client.h"
+#include "ircd.h"
+#include "logger.h"
+#include "newconf.h"
+#include "numeric.h"
+#include "send.h"
+#include "s_newconf.h"
+#include "s_serv.h"
+#include "s_user.h"
+
+#define IsOperQuarantine(x) (HasPrivilege((x), "oper:quarantine"))
+#define IsQuarantined(x) ((x)->umodes & user_modes['q'])
+#define DEFAULT_JOIN_REASON "Cannot join channel (usermode +q) - you need to be logged into your NickServ account"
+#define DEFAULT_MSG_REASON "Cannot send to nick/channel (usermode +q) - you need to be logged into your NickServ account"
+#define DEFAULT_OTHER_MSG_REASON "Cannot send to user - they are quarantined and will be unable to respond to you"
+#define DEFAULT_APPLY_MSG "You have been quarantined and must log into your NickServ account before you can join channels. Please see /STATS p for assistance."
+#define DEFAULT_REMOVE_MSG "You are no longer quarantined and can freely join channels."
+
+static const char quarantine_desc[] = "QUARANTINE system and umode +q to restrict unidentified connections.";
+static char *join_reason = NULL;
+static char *msg_reason = NULL;
+static char *other_msg_reason = NULL;
+static char *apply_msg = NULL;
+static char *remove_msg = NULL;
+static int part_channels = 1;
+static char **allowed_channels = NULL;
+
+static struct Client *allowed_umode_change = NULL;
+
+static int modinit(void);
+static void moddeinit(void);
+static void me_quarantine(struct MsgBuf *, struct Client *, struct Client *, int, const char *[]);
+static void mo_quarantine(struct MsgBuf *, struct Client *, struct Client *, int, const char *[]);
+static void me_unquarantine(struct MsgBuf *, struct Client *, struct Client *, int, const char *[]);
+static void mo_unquarantine(struct MsgBuf *, struct Client *, struct Client *, int, const char *[]);
+static void add_quarantine(struct Client *source_p, struct Client *target_p, const char *reason);
+static void remove_quarantine(struct Client *source_p, struct Client *target_p);
+static void quarantine_account_change(void *);
+static void quarantine_change_umode(void *);
+static void quarantine_conf_info(void *);
+static void quarantine_default_umode_begin(void *);
+static void quarantine_default_umode_end(void *);
+static void quarantine_join_channel(void *);
+static void quarantine_privmsg_channel(void *);
+static void quarantine_privmsg_user(void *);
+static void quarantine_reset_conf(void *);
+static void quarantine_set_allow_channels(void *);
+
+struct Message quarantine_msgtab = {
+	"QUARANTINE", 0, 0, 0, 0,
+	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_quarantine, 3 }, { mo_quarantine, 3 } }
+};
+
+struct Message unquarantine_msgtab = {
+	"UNQUARANTINE", 0, 0, 0, 0,
+	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_unquarantine, 2 }, { mo_unquarantine, 2 } }
+};
+
+struct ConfEntry conf_quarantine_table[] = {
+	{ "allow_channels", CF_QSTRING | CF_FLIST, quarantine_set_allow_channels, 0, NULL },
+	{ "apply_msg", CF_QSTRING, NULL, BUFSIZE, &apply_msg },
+	{ "join_reason", CF_QSTRING, NULL, BUFSIZE, &join_reason },
+	{ "msg_reason", CF_QSTRING, NULL, BUFSIZE, &msg_reason },
+	{ "other_msg_reason", CF_QSTRING, NULL, BUFSIZE, &other_msg_reason },
+	{ "part_channels", CF_YESNO, NULL, 0, &part_channels },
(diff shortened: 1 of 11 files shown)


extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+static void quarantine_default_umode_end(void *);
+static void quarantine_join_channel(void *);
+static void quarantine_privmsg_channel(void *);
+static void quarantine_privmsg_user(void *);
+static void quarantine_reset_conf(void *);
+static void quarantine_set_allow_channels(void *);
+
+struct Message quarantine_msgtab = {
+	"QUARANTINE", 0, 0, 0, 0,
+	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_quarantine, 2 }, { mo_quarantine, 2 } }
```
dwfreed:
```suggestion
	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_quarantine, 3 }, { mo_quarantine, 3 } }
```

If you set `min_para` correctly, you don't have to check the parameters at all yourself.  This number is checked against `parc`, hence why it must be one more than the number of parameters; `parc` is always at least `1`, and `parv[0]` is the command.  Core additionally checks that `parv[min_para - 1]` is not an empty string, and in order to have 2 parameters, the first parameter automatically cannot be an empty string either.

extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+};
+
+struct Message unquarantine_msgtab = {
+	"UNQUARANTINE", 0, 0, 0, 0,
+	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_unquarantine, 1 }, { mo_unquarantine, 1 } }
```
dwfreed:
```suggestion
	{ mg_unreg, mg_not_oper, mg_not_oper, mg_ignore, { me_unquarantine, 2 }, { mo_unquarantine, 2 } }
```

Likewise

extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+#include "numeric.h"
+#include "send.h"
+#include "s_newconf.h"
+#include "s_serv.h"
+#include "s_user.h"
+
+#define IsOperQuarantine(x) (HasPrivilege((x), "oper:quarantine"))
+#define IsQuarantined(x) ((x)->umodes & user_modes['q'])
+#define DEFAULT_JOIN_REASON "Cannot join channel (+q) - you need to be logged into your NickServ account"
+#define DEFAULT_MSG_REASON "Cannot send to nick/channel (+q) - you need to be logged into your NickServ account"
```
dwfreed:
```suggestion
#define DEFAULT_JOIN_REASON "Cannot join channel (usermode +q) - you need to be logged into your NickServ account"
#define DEFAULT_MSG_REASON "Cannot send to nick/channel (usermode +q) - you need to be logged into your NickServ account"
```

These should call out that it's a usermode, as the mode character present in other forms of these numerics is a channel mode

extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+		{
+			if (!irccmp(data->chptr->chname, allowed_channels[i]))
+				return;
+		}
+	}
+
+	data->approved = ERR_CANNOTSENDTOCHAN;
+
+	/* Don't give error messages for TAGMSG since many clients autogenerate these (e.g. +typing) */
+	if (data->msgtype != MESSAGE_TYPE_TAGMSG)
```
dwfreed:
```suggestion
	if (data->msgtype == MESSAGE_TYPE_PRIVMSG)
```

Notices should not trigger numerics either

extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+
+	/* if some other hook function already rejected this attempt, keep that rejection */
+	if (data->approved != 0)
+		return;
+
+	if (!IsQuarantined(data->source_p))
+	{
+		if (IsQuarantined(data->target_p) && !IsOper(data->source_p) && !IsService(data->source_p))
+		{
+			if (data->msgtype != MESSAGE_TYPE_TAGMSG)
```
dwfreed:
```suggestion
			if (data->msgtype == MESSAGE_TYPE_PRIVMSG)
```

Likewise

extensions/m_quarantine.c
```diff
@@ -0,0 +1,601 @@
...
+	}
+
+	/* let quarantined users message opers and services */
+	if (IsOper(data->target_p) || IsService(data->target_p))
+		return;
+
+	data->approved = ERR_CANNOTSENDTOCHAN;
+
+	/* Don't give error messages for TAGMSG since many clients autogenerate these (e.g. +typing) */
+	if (data->msgtype != MESSAGE_TYPE_TAGMSG)
```
dwfreed:
```suggestion
	if (data->msgtype == MESSAGE_TYPE_PRIVMSG)
```

Likewise

modules/m_services.c
```diff
@@ -131,6 +139,8 @@ me_su(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p,
 	if(!target_p->user)
 		return;
 
+	strncpy(old_suser, target_p->user->suser, sizeof(old_suser) - 1);
```
dwfreed:
```suggestion
	rb_strlcpy(old_suser, target_p->user->suser, sizeof(old_suser));
```

`strncpy` is ***VERBOTEN*** in ircd code

modules/m_services.c
```diff
@@ -151,16 +161,22 @@ me_su(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p,
...
 
 static void
 me_login(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p,
 	int parc, const char *parv[])
 {
+	char old_suser[NICKLEN + 1] = { 0 };
 	if(!IsPerson(source_p))
 		return;
 
+	strncpy(old_suser, source_p->user->suser, sizeof(old_suser) - 1);
```
dwfreed:
```suggestion
	rb_strlcpy(old_suser, source_p->user->suser, sizeof(old_suser));
```

Likewise

modules/m_signon.c
```diff
@@ -406,10 +412,12 @@ send_signon(struct Client *client_p, struct Client *target_p,
 		const char *nick, const char *user, const char *host,
 		unsigned int newts, const char *login)
 {
+	char old_suser[NICKLEN + 1] = { 0 };
 	sendto_server(client_p, NULL, CAP_TS6, NOCAPS, ":%s SIGNON %s %s %s %ld %s",
 			use_id(target_p), nick, user, host,
 			(long) target_p->tsinfo, *login ? login : "0");
 
+	strncpy(old_suser, target_p->user->suser, sizeof(old_suser) - 1);
```
dwfreed:
```suggestion
	rb_strlcpy(old_suser, target_p->user->suser, sizeof(old_suser));
```

Likewise

dwfreed (dismissed):

extensions/m_quarantine.c
```diff
@@ -0,0 +1,577 @@
...
+#include "ircd.h"
+#include "logger.h"
+#include "newconf.h"
+#include "numeric.h"
+#include "send.h"
+#include "s_newconf.h"
+#include "s_serv.h"
+#include "s_user.h"
+
+#define IsOperQuarantine(x) (HasPrivilege((x), "oper:quarantine"))
```
aaronmdjones, line 37: This should use `MayHavePrivilege` so that opers on non-solanum servers (e.g. services) can use it.

extensions/m_quarantine.c
```diff
@@ -0,0 +1,577 @@
...
+{
+	struct Client *target_p = find_named_person(parv[1]);
+	if (target_p == NULL)
+	{
+		sendto_one_numeric(source_p, ERR_NOSUCHNICK, form_str(ERR_NOSUCHNICK), parv[1]);
+		return;
+	}
+
+	if (!MyClient(target_p))
+		return;
```
aaronmdjones: Invert the conditional and drop the return

```C
if (MyClient(target_p))
    add_quarantine(source_p, target_p, parv[2]);
```

extensions/m_quarantine.c
```diff
@@ -0,0 +1,577 @@
...
```
aaronmdjones: Likewise

extensions/m_quarantine.c
skizzerz: `IsOperQuarantine` is only called in the oper handlers which only runs for local opers. We don't do priv checking in the encap handlers. So, services or other non-solanum linked servers should be unimpacted?

aaronmdjones: Ah, yes, good point.

extensions/m_quarantine.c
```diff
@@ -0,0 +1,577 @@
...
+			data->approved = ERR_CANNOTSENDTOUSER;
+		}
+		return;
+	}
+
+	/* let quarantined users message opers and services */
+	if (IsOper(data->target_p) || IsService(data->target_p))
+		return;
+
+	data->approved = ERR_CANNOTSENDTOCHAN;
```
aaronmdjones, line 524: Shouldn't this be `ERR_CANNOTSENDTOUSER` to match the block above?

Likewise with the if block below.

extensions/m_quarantine.c
```diff
@@ -0,0 +1,573 @@
...
+		rb_free(join_reason);
+
+	if (msg_reason != NULL)
+		rb_free(msg_reason);
+
+	if (other_msg_reason != NULL)
+		rb_free(other_msg_reason);
+
+	if (remove_msg != NULL)
+		rb_free(remove_msg);
```
aaronmdjones: `rb_free(NULL)` is safe; no need to test if they're `NULL`

aaronmdjones (dismissed):

aaronmdjones (approved):

PR #495, opened 2026-02-03
Add support for building with meson instead of autotools

progval:
It's significantly faster, primarily because the autotools setup links every plugin every time in a non-parallelizable for loop.

clean build with autotools:

```
> time ./configure --prefix /tmp/solanum-autotools/
[...]
Executed in   19.20 secs    fish           external
   usr time   12.98 secs    2.06 millis   12.98 secs
   sys time    6.60 secs    0.78 millis    6.60 secs
> time make -j 16
[...]
Executed in   25.57 secs    fish           external
   usr time  111.30 secs    0.00 millis  111.30 secs
   sys time   37.90 secs    2.43 millis   37.90 secs
> time make -j 16 install
[...]
Executed in   43.56 secs    fish           external
   usr time   35.25 secs    1.71 millis   35.25 secs
   sys time   18.99 secs    0.00 millis   18.99 secs
```

(total wall clock: 88s)

incremental build&install (with no change):

```
> time make -j 16 install
Executed in   43.65 secs    fish           external
   usr time   35.09 secs    0.00 millis   35.09 secs
   sys time   19.18 secs    2.13 millis   19.17 secs
```

clean build with meson:

```
> time meson setup --prefix /tmp/solanum-meson/ --optimization 2 build/
[...]
Executed in    3.16 secs    fish           external
   usr time    2.58 secs    0.00 millis    2.58 secs
   sys time    0.57 secs    1.80 millis    0.57 secs
> time meson compile -C build/
[...]
Executed in   12.33 secs    fish           external
   usr time   98.92 secs    0.00 millis   98.92 secs
   sys time   20.02 secs    2.31 millis   20.02 secs
> time meson install -C build/
Executed in    1.08 secs    fish           external
   usr time    1.08 secs    2.58 millis    1.07 secs
   sys time    0.12 secs    0.00 millis    0.12 secs
```

(total wall clock: 17s)

incremental build&install (with no change):

```
> time meson install -C build/
Executed in    1.40 secs    fish           external
   usr time    1.34 secs    1.75 millis    1.34 secs
   sys time    0.19 secs    0.65 millis    0.19 secs
```


.github/workflows/ci.yml
```diff
@@ -23,10 +23,10 @@ on:
       - 'README.md'
 
 jobs:
-  linux:
+  linux-autotools:
     name: Linux
     runs-on: ${{ matrix.os }}
-    strategy:
+    strategy: &strategy
```
dwfreed: You give the strategy an anchor, but you do not use it

help/meson.build
```diff
@@ -0,0 +1,73 @@
...
+      fi
+    done
+
+    echo 'Help topics available to opers:' > opers/index
+    echo '' >> opers/index
+    cat opers/index.tmp | sed -e 's|^opers/||' | sort -u | tr a-z A-Z | column -c 65 -s ' ' -x | expand >> opers/index
+    rm -f opers/index.tmp
+
+    touch $OUTPUT0 $OUTPUT1
+    '''
```
dwfreed: Split this out into a script

help/meson.build
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  build_by_default: true,
+  build_always_stale: true
```
dwfreed: Can we find a way to make this not required? Having to rebuild the help index to run the tests seems rather broken

librb/meson.build
```diff
@@ -0,0 +1,148 @@
...
+  dependencies: librb_deps
+)
+
+pkg = import('pkgconfig')
+pkg.generate(librb,
+  name: 'librb',
+  description: 'libratbox',
+  subdirs: 'librb',
+  extra_cflags: ['-I${includedir}/librb']
+)
```
dwfreed: Ditch this

librb/meson.build
```diff
@@ -0,0 +1,148 @@
...
+
+librb_inc = include_directories('include', '.', is_system: false)
+
+librb_build_inc = include_directories('.', is_system: false)
+
+librb = shared_library('rb',
+  [librb_sources, librb_version_c, librb_dash_config_h, librb_underscore_config_h, librb_serno_h],
+  dependencies: librb_deps,
+  include_directories: [librb_inc, librb_build_inc],
+  c_args: ['-I' + meson.current_build_dir()],
```
dwfreed: Why is this not just in `include_directories`? Or automatically handled by `implicit_include_directories`?

ircd/meson.build
```diff
@@ -0,0 +1,94 @@
...
+  'send.c',
+  'snomask.c',
+  'sslproc.c',
+  'substitution.c',
+  'supported.c',
+  'tgchange.c',
+  'whowas.c',
+)
+
+libircd_inc = include_directories('../include', '../librb/include')
```
dwfreed: Should not need to include librb here as it should be handled by the librb dep?

ircd/meson.build
```diff
@@ -0,0 +1,94 @@
...
+
+libircd = shared_library('ircd',
+  [libircd_sources, ircd_parser, ircd_lexer, ircd_version, serno_h],
+  dependencies: [librb_dep, ltdl_dep, sqlite3_dep],
```
dwfreed: sqlite dep should be confined to the bandb directory

modules/meson.build
```diff
@@ -0,0 +1,124 @@
+module_inc = include_directories('../include', '../librb/include')
```
dwfreed: Should be completely unnecessary due to handling by libircd and librb deps?

extensions/meson.build
```diff
@@ -0,0 +1,101 @@
+extension_inc = include_directories('../include', '../librb/include')
```
dwfreed: Should be completely unnecessary due to handling by the libircd and librb deps?

tests/meson.build
```diff
@@ -0,0 +1,61 @@
...
+
+test_util_sources = files(
+  'ircd_util.c',
+  'client_util.c',
+)
+
+test_utils = static_library('test_utils',
+  [test_util_sources, serno_h],
+  dependencies: [libircd_dep, librb_dep],
+  include_directories: [include_directories('../include', '../librb/include'), include_directories('..')]
```
dwfreed: include and librb/include should be handled by the deps

tests/meson.build
```diff
@@ -0,0 +1,61 @@
...
+  'serv_connect1': 'serv_connect1.c',
+  'substitution1': 'substitution1.c',
+}
+
+foreach test_name, test_source : test_programs
+  test_exe = executable(test_name,
+    test_source,
+    dependencies: [libircd_dep, librb_dep, dl_dep],
+    link_with: [test_utils, tap_lib],
+    include_directories: [include_directories('../include', '../librb/include'), include_directories('..')],
```
dwfreed: Likewise

tests/meson.build
```diff
@@ -0,0 +1,61 @@
...
+  )
+
+  test(test_name, test_exe)
+endforeach
+
+runtests = executable('runtests',
+  'runtests.c',
+  c_args: [
+    '-DC_TAP_SOURCE="' + meson.project_source_root() + '/tests"',
+    '-DC_TAP_BUILD="' + meson.project_build_root() + '/tests"',
```
dwfreed: These should use the `current` variants instead of manually building the path

tools/meson.build
```diff
@@ -0,0 +1,15 @@
+tools_inc = include_directories('../include', '../librb/include')
```
dwfreed: librb/include should be able to be dropped due to dep handling?

meson.build
```diff
@@ -0,0 +1,398 @@
...
+
+if get_option('openssl').enabled() and not have_openssl
+  error('OpenSSL requested but not found')
+endif
+if get_option('mbedtls').enabled() and not have_mbedtls
+  error('mbedTLS requested but not found')
+endif
+if get_option('gnutls').enabled() and not have_gnutls
+  error('GnuTLS requested but not found')
+endif
```
dwfreed: The various `required` values should just be the respective option's `enabled()` value

meson.build
```diff
@@ -0,0 +1,398 @@
...
+
+# For compat with autotools. See https://www.gnu.org/software/libtool/manual/html_node/Autoconf-macros.html#index-LT_005fSYS_005fMODULE_005fEXT
+if host_machine.system() == 'windows'
+  conf_data.set_quoted('LT_MODULE_EXT', '.dll')
+elif host_machine.system() == 'darwin'
+  conf_data.set_quoted('LT_MODULE_EXT', '.so')
+else
+  conf_data.set_quoted('LT_MODULE_EXT', '.so')
+endif
```
dwfreed: Ditch the system tests, we only care about Linux.

meson.build
```diff
@@ -0,0 +1,398 @@
...
+        exit 0
+      fi
+    fi
+    echo '#define SERNO "unknown"'
+    echo '#define DATECODE 0UL'
+    '''
+  ],
+  install: false,
+  build_by_default: true,
+  build_always_stale: true
```
dwfreed: Also fix this to not always be stale

meson.build
```diff
@@ -0,0 +1,398 @@
...
+      datecode=$(git rev-list -1 --no-commit-header --format=%ct HEAD 2>/dev/null || true)
+      if [ -n "$revh" ] && [ -n "$datecode" ]; then
+        echo "#define SERNO \\"$revh\\""
+        echo "#define DATECODE ${datecode}UL"
+        exit 0
+      fi
+    fi
+    echo '#define SERNO "unknown"'
+    echo '#define DATECODE 0UL'
+    '''
```
dwfreed: Split this out into a script

meson.build
```diff
@@ -0,0 +1,398 @@
...
+endif
+
+if get_option('asan')
+  if cc.has_argument('-fsanitize=address')
+    c_args += ['-fsanitize=address']
+    link_args += ['-fsanitize=address']
+  endif
+endif
+
+rpath_arg = '-Wl,-rpath,' + prefix / get_option('libdir')
```
dwfreed: This should be done with `install_rpath`

meson.build
```diff
@@ -0,0 +1,398 @@
...
+link_args += [rpath_arg]
+
+add_project_arguments(c_args, language: 'c')
+add_project_link_arguments(link_args, language: 'c')
+add_project_arguments('-D_GNU_SOURCE', language: 'c')
+add_project_arguments('-I' + meson.current_build_dir(), language: 'c')
```
dwfreed: This should be unnecessary

ircd/meson.build
```diff
@@ -0,0 +1,94 @@
...
+  include_directories: libircd_inc,
+  install: true,
+  install_dir: get_option('libdir')
+)
+
+libircd_dep = declare_dependency(
+  link_with: libircd,
+  include_directories: libircd_inc,
+  dependencies: [librb_dep, ltdl_dep, sqlite3_dep]
```
dwfreed: sqlite dep should be confined to the bandb directory

dwfreed (changes requested):
Good start, a lot of thoughts

help/meson.build
progval: We would need to list all the help topics in meson.build, and I'm not sure it would be rebuilt if we removed a topic.

meson.build
```diff
@@ -0,0 +1,398 @@
...
+    link_args += ['-pg']
+  endif
+endif
+
+if get_option('asan')
+  if cc.has_argument('-fsanitize=address')
+    c_args += ['-fsanitize=address']
+    link_args += ['-fsanitize=address']
+  endif
+endif
```
dwfreed: Ditch this in favor of meson's

help/meson.build
dwfreed: One thought is that this could be an install script?

README.md
```diff
@@ -46,26 +37,28 @@ These are known issues and workarounds for various platforms.
...
+   * OpenSSL 1.0.0 or newer (`-Dopenssl=enabled`)
+   * LibreSSL (`-Dopenssl=enabled`)
+   * mbedTLS (`-Dmbedtls=enabled`)
+   * GnuTLS (`-Dgnutls=enabled`)
+
+   If multiple TLS libraries are available, the build system will prefer mbedTLS > OpenSSL > GnuTLS.
+   Use `=enabled` instead of `=auto` to force a specific library.
 
- * For certificate-based oper CHALLENGE, OpenSSL 1.0.0 or newer.
+ * For certificate-based oper CHALLENGE, OpenSSL 0.9.7 or newer.
```
aaronmdjones: This is not correct (neither is the original).

OpenSSL < 1.1.0 lacks the `EVP_MD_CTX_new()` function used in `modules/challenge.c:322`.

It should be corrected to say 1.1.0 or newer.

meson.build
progval: that's how it finds `setup.h`

dwfreed: We should put `setup.h` in a less dumb place then, like in `include` and/or use a proper `include_directories` for it.  Manually setting an include dir on the whole project seems... bad.

meson.build
progval: How? Make it depend on `.git/HEAD` somehow?

meson.build
progval: Like this?

```diff
diff --git a/authd/meson.build b/authd/meson.build
index c53a3933..1cb75820 100644
--- a/authd/meson.build
+++ b/authd/meson.build
@@ -16,5 +16,6 @@ authd = executable('authd',
   dependencies: [librb_dep],
   include_directories: include_directories('../include'),
   install: true,
-  install_dir: pkglibexecdir
+  install_dir: pkglibexecdir,
+  install_rpath: rpath,
 )
diff --git a/bandb/meson.build b/bandb/meson.build
index 311d47bd..51503632 100644
--- a/bandb/meson.build
+++ b/bandb/meson.build
@@ -8,7 +8,8 @@ bandb = executable('bandb',
   dependencies: [librb_dep, sqlite3_dep],
   include_directories: include_directories('../include'),
   install: true,
-  install_dir: pkglibexecdir
+  install_dir: pkglibexecdir,
+  install_rpath: rpath,
 )
 
 bantool = executable('solanum-bantool',
@@ -16,5 +17,6 @@ bantool = executable('solanum-bantool',
   dependencies: [librb_dep, sqlite3_dep],
   include_directories: include_directories('../include'),
   install: true,
-  install_dir: get_option('bindir')
+  install_dir: get_option('bindir'),
+  install_rpath: rpath,
 )
diff --git a/ircd/meson.build b/ircd/meson.build
index 6161224a..fc23f241 100644
--- a/ircd/meson.build
+++ b/ircd/meson.build
@@ -76,7 +76,8 @@ libircd = shared_library('ircd',
   dependencies: [librb_dep, ltdl_dep],
   include_directories: libircd_inc,
   install: true,
-  install_dir: get_option('libdir')
+  install_dir: get_option('libdir'),
+  install_rpath: rpath,
 )
 
 libircd_dep = declare_dependency(
@@ -90,5 +91,6 @@ solanum = executable('solanum',
   dependencies: [libircd_dep],
   export_dynamic: true,
   install: true,
-  install_dir: get_option('bindir')
+  install_dir: get_option('bindir'),
+  install_rpath: rpath,
 )
diff --git a/meson.build b/meson.build
index 6ef41fa2..2c22b220 100644
--- a/meson.build
+++ b/meson.build
@@ -298,11 +298,9 @@ if get_option('profile')
   endif
 endif
 
-rpath_arg = '-Wl,-rpath,' + prefix / get_option('libdir')
-link_args += [rpath_arg]
+rpath = prefix / get_option('libdir')
 
 add_project_arguments(c_args, language: 'c')
-add_project_link_arguments(link_args, language: 'c')
 add_project_arguments('-D_GNU_SOURCE', language: 'c')
 
 subdir('include')
diff --git a/ssld/meson.build b/ssld/meson.build
index d43f37d8..736237a5 100644
--- a/ssld/meson.build
+++ b/ssld/meson.build
@@ -3,5 +3,6 @@ ssld = executable('ssld',
   dependencies: [librb_dep],
   include_directories: include_directories('../include'),
   install: true,
-  install_dir: pkglibexecdir
+  install_dir: pkglibexecdir,
+  install_rpath: rpath,
 )
diff --git a/tests/meson.build b/tests/meson.build
index 2c0d1588..214eec77 100644
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -57,5 +57,6 @@ runtests = executable('runtests',
   ],
   link_with: [tap_lib],
   include_directories: [include_directories('tap'), include_directories('..')],
-  build_by_default: true
+  build_by_default: true,
+  install_rpath: rpath,
 )
diff --git a/tools/meson.build b/tools/meson.build
index 0e10ea9c..0f41a183 100644
--- a/tools/meson.build
+++ b/tools/meson.build
@@ -10,6 +10,7 @@ foreach name, deps : tools
     dependencies: deps,
     include_directories: tools_inc,
     install: true,
-    install_dir: get_option('bindir')
+    install_dir: get_option('bindir'),
+    install_rpath: rpath,
   )
 endforeach
```

dwfreed: Looks right to me, :shipit:

help/meson.build
progval: That was a good though.

bandb/meson.build
```diff
@@ -0,0 +1,22 @@
...
+  install_dir: pkglibexecdir,
+  install_rpath: rpath,
+)
+
+bantool = executable('solanum-bantool',
+  ['bantool.c'] + bandb_common_sources,
+  dependencies: [librb_dep, sqlite3_dep],
+  include_directories: include_directories('../include'),
+  install: true,
+  install_dir: get_option('bindir'),
```
eli-schwartz: The installation directory for an executable is already implicitly bindir, you do not need to override it.

help/make_index.SH
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+SOURCE="$1"
+OUTPUT="$2"
+
+USERS_INDEX=$OUTPUT/users/index
+OPERS_INDEX=$OUTPUT/opers/index
+
+# Build users index
+mkdir -p $OUTPUT/users/
```
eli-schwartz: Quote variables to prevent this from breaking when building in a directory that has spaces. :)

Meson does support this well. We make sure of it by running the entire meson testsuite in e.g.

```
meson.git/test cases/common/1 test foo works/
meson.git/test cases/common/2 test bar works/
```

:P

help/make_index.SH
```diff
@@ -0,0 +1,40 @@
...
+done
+
+# Add symlinks to users index
+for link in topic accept cmode admin names links away whowas version kick who invite quit join list nick oper part time credits motd userhost users whois ison lusers user help pass error challenge knock ping pong map trace chantrace extban monitor batch; do
+  echo $(basename "$link") >> $USERS_INDEX.tmp
+done
+
+echo 'Help topics available to users:' > $USERS_INDEX
+echo '' >> $USERS_INDEX
+cat $USERS_INDEX.tmp | sed -e 's|^users/||' | sort -u | tr a-z A-Z | column -c 65 -x | expand >> $USERS_INDEX
```
eli-schwartz: "useless use of cat" -- can pass input to sed via redirection. Admittedly this is copying existing code as-is to a new location...

help/meson.build
```diff
@@ -0,0 +1,33 @@
+# Generate help index files
+meson.add_install_script(
+  'sh',
+  '-c',
+  'helpdir=${DESTDIR}/$0 && echo Installing help indexes to ${helpdir} && ${MESON_SOURCE_ROOT}/help/make_index.SH ${MESON_SOURCE_ROOT}/help ${helpdir}',
```
eli-schwartz: Can't you just add $DESTDIR support directly to the script itself?

ircd/meson.build
```diff
@@ -0,0 +1,96 @@
...
+)
+
+libircd_inc = include_directories('../include')
+
+libircd = shared_library('ircd',
+  [libircd_sources, ircd_parser, ircd_lexer, ircd_version, serno_h],
+  dependencies: [librb_dep, ltdl_dep],
+  include_directories: libircd_inc,
+  install: true,
+  install_dir: get_option('libdir'),
```
eli-schwartz: Defaults to libdir already.

librb/meson.build
```diff
@@ -0,0 +1,139 @@
...
+  output: 'version.c',
+  configuration: librb_version_conf
+)
+
+# Create a copy of serno.h in librb's include directory for version.c
+librb_include_dir = meson.current_build_dir() / 'include'
+librb_serno_h = custom_target('librb_serno.h',
+  input: serno_h,
+  output: 'serno_librb.h',
+  command: ['sh', '-c', 'mkdir -p ' + librb_include_dir + ' && mkdir -p $(dirname "$1") && cp $0 ' + librb_include_dir + '/serno.h && cp $0 $1', '@INPUT@', '@OUTPUT@'],
```
eli-schwartz: Why is this copying the output to somewhere other than the current build dir?

librb/meson.build
```diff
@@ -0,0 +1,139 @@
...
+# For compat with autotools mess
+librb_underscore_config_h = configure_file(
+  input: librb_dash_config_h,
+  output: 'librb_config.h',
+  copy: true
+)
+
+librb_deps = [socket_dep] + tls_deps
+if rt_dep.found()
+  librb_deps += [rt_dep]
```
eli-schwartz: It's okay to unconditionally add a not-found-dependency. If it is not found, it will evaluate to "don't add any compile or link flags".

meson.build
```diff
@@ -0,0 +1,348 @@
...
+)
+
+cc = meson.get_compiler('c')
+
+# Required dependencies
+flex = find_program('flex', 'lex', required: true)
+bison = find_program('bison', 'yacc', 'byacc', required: true)
+socket_dep = cc.find_library('socket', required: false)
+sqlite3_dep = dependency('sqlite3', required: true)
+dl_dep = cc.find_library('dl', required: false)
```
eli-schwartz: Meson 0.62 would allow you to use 
https://mesonbuild.com/Dependencies.html#dl-libdl

meson.build
```diff
@@ -0,0 +1,348 @@
...
+    have_hyperscan = true
+  endif
+endif
+
+# SCTP
+sctp_dep = dependency('', required: false)
+have_sctp = false
+if get_option('sctp').allowed()
+  sctp_lib = cc.find_library('sctp', required: get_option('sctp'))
+  if sctp_lib.found() and cc.has_header('netinet/sctp.h')
```
eli-schwartz: What's the intended logic flow here?

If passing sctp=disabled, you can just do:
```meson
sctp_lib = cc.find_library('sctp', required: get_option('sctp'))
```

and sctp_lib.found() will evaluate to false. Meson will log a message that it is NOT searching for this library, "because feature *sctp* is *disabled*"

If sctp=enabled then the lack of the library will be a fatal configure error. However, if the library exists but the relevant header does not exist, then have_sctp will be set to false and the functionality is silently skipped?

eli-schwartz: Additionally, you may find it useful to pass `has_headers: ['netinet/sctp.h']` as a parameter to the find_library call, as this allows you to handle both the library lookup and the header lookup as a single unit which the `required:` kwarg then applies to

meson.build
```diff
@@ -0,0 +1,348 @@
...
+tls_backend = 'none'
+tls_deps = []
+have_openssl = false
+have_mbedtls = false
+have_gnutls = false
+
+# mbedTLS
+if get_option('mbedtls').allowed()
+  mbedtls_required = get_option('mbedtls').enabled()
+  mbedtls_lib = cc.find_library('mbedtls', required: mbedtls_required)
```
eli-schwartz: This seems very needlessly complex. The entire point of a feature option is to pass it directly to `required`. You shouldn't be manually checking the .enabled() or .allowed() state.

```meson
mbedtls_lib = cc.find_library('mbedtls', required: get_option('mbedtls'))
mbedx509_lib = cc.find_library('mbedx509', required: get_option('mbedtls'))
mbedcrypto_lib = cc.find_library('mbedcrypto', required: get_option('mbedtls'))

# at this point if mbedtls is not allowed, all will be not-found.
# if it is enabled, and any were missing, meson has already aborted
# 
# now let's check: if it is auto, were (all of them) found?
if mbedtls_lib.found() and mbedx509_lib.found() and mbedcrypto_lib.found()
```

meson.build
```diff
@@ -0,0 +1,348 @@
...
+  conf_data.set('CUSTOM_BRANDING', 1)
+endif
+
+# Program prefix
+program_prefix = get_option('program_prefix')
+conf_data.set_quoted('PROGRAM_PREFIX', program_prefix)
+
+# Features
+if get_option('oper_chghost')
+  conf_data.set('ENABLE_OPER_CHGHOST', 1)
```
eli-schwartz, line 147: set10 may be convenient here. What's the behavior of the code if the define is set to 0?

meson.build
```diff
@@ -0,0 +1,348 @@
...
+  conf_data.set('HAVE_MBEDTLS', 1)
+endif
+if have_gnutls
+  conf_data.set('HAVE_GNUTLS', 1)
+endif
+
+# Performance
+assert_level = get_option('assert')
+if assert_level == 'none'
+  conf_data.set('NDEBUG', 1)
```
eli-schwartz: There is a `b_ndebug` option built into meson, for the record.

meson.build
```diff
@@ -0,0 +1,348 @@
...
+# We only support Linux, though
+conf_data.set_quoted('LT_MODULE_EXT', '.so')
+
+
+c_args = []
+link_args = []
+
+if get_option('warnings')
+  c_args += [
+    '-Wall',
```
eli-schwartz: -Wall -Wextra are built into meson as `meson setup -Dwarning_level=2` (can go -pedantic too if you like, and there's a hand-rolled implementation of clang -Weverything for GCC, even).

meson.build
dwfreed: Bad:

https://github.com/solanum-ircd/solanum/blob/59e35e3f5e0943729f2b55f860416b63c4a0e73f/modules/m_chghost.c#L202

eli-schwartz: RIP...

Stuff like this is always annoying for any port. :D It takes a lot of close scrutiny to make sure that something which isn't defined is undefined because the build system detected it as disabled vs the build system forgot to handle it, or handled it in a never-executed branch.

But of course refactoring this is quite outside the scope of a build system change.

dwfreed: I'm all for refactoring this to make the build system change suck less.  And it's a simple change, `#ifdef` -> `#if`

meson.build
progval: @dwfreed What meson version should we target? fwiw, Debian 11 has 0.56 (which we don't support anyway) and Debian 12 has 1.0.1.

help/make_index.SH
progval: heh, I like it this way because it can be read from left to right (file -> "sed" -> regexp -> sort -> ..., instead of "sed" -> file -> regexp -> sort). That's up to the maintainers though.

eli-schwartz:
```
<"${file}" sed "s/reg/exp/" | sort -u | ...
```

:) :)

meson.build
dwfreed: I'd say bumping up one more version to 0.62 wouldn't hurt anything.  If we decide to only support the current LTS-equivalent release of the major distros, the limiting factor becomes Ubuntu 24.04, with meson 1.3.

help/make_index.SH
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 
-SOURCE="$1"
-OUTPUT="$2"
+echo "Writing help indexes to $OUTPUT from $INPUT"
```
eli-schwartz: I think you still do need the assignment for $1 and $2, though

help/meson.build
```diff
@@ -2,7 +2,7 @@
 meson.add_install_script(
   'sh',
   '-c',
-  'helpdir=${DESTDIR}/$0 && echo Installing help indexes to ${helpdir} && ${MESON_SOURCE_ROOT}/help/make_index.SH ${MESON_SOURCE_ROOT}/help ${helpdir}',
+  f'SOURCE="${MESON_SOURCE_ROOT}/help" OUTPUT="${DESTDIR}/$0" ${MESON_SOURCE_ROOT}/help/make_index.SH',
```
eli-schwartz: ... Because my suggestion really was that you could skip the sh -c and just do:
```meson
meson.add_install_script(files('make_index.SH'), meson.current_source_dir(), helpdir)
```

And rely on the shell script detecting when it is run with DESTDIR set.

help/make_index.SH
progval: TIL

help/meson.build
eli-schwartz: `meson install` will log the command that gets run, so this (avoiding sh -c) also has a readability advantage.

meson.build
dwfreed: Note that bumping to 0.62 will require dropping 22.04 from CI (which only has 0.61), which is fine by me.

help/meson.build
progval: nice, thanks

help/make_index.SH
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+SOURCE="$1"
+OUTPUT="$DESTDIR$2"
```
dwfreed:
```suggestion
OUTPUT="${DESTDIR}$2"
```

But note this will do the wrong thing in the autotools case when DESTDIR is set.

progval: how so?

dwfreed: DESTDIR doesn't have to end in a /, so if it's set to `some/path`, with what you're passing as param 2 from Makefile.am, you get `some/path./`; and even with a /, you get `some/path/./` which will also be wrong.

eli-schwartz: Doubling up the forward slashes is fine (`$DESTDIR/$2` correctly handles the case where $2 starts with a slash).

The component `./` is a no-op.

In both cases if the resulting path is data that gets written to some output file then it is "ugly", but if it's just being used as a directive for where to install some files, it really does not matter.

...

Generally speaking one should not define DESTDIR at compile time, only at install time. Still, I suppose one could `./configure && DESTDIR=.... make install` without an intervening build stage, which... Meh. :D 

Only processing DESTDIR if `[ "$1" = --install ]` could handle that I suppose. Given that the script has two modes, one to run at build time and one to run at install time.

eli-schwartz: I should perhaps note at this time that for portability with Windows (I know, you don't support it and this is a POSIX shell script anyway), meson does guarantee that $MESON_INSTALL_DESTDIR_PREFIX is set to `$DESTDIR` + `get_option('prefix')` with allowance made for the latter being `C:/MyProg/` (windows drives cannot have a directory naively prepended, it needs to make the root rather than the drive be DESTDIR-relative).

That means you could pass a relative rather than an absolute path as "helpdir", and rely on the fact that only meson will ever define the `MESON_INSTALL_DESTDIR_PREFIX` variable (and that only during installation).

meson.build
progval: there are almost 200 #ifdef, can we keep this change out of this PR?

dwfreed: You don't have to change all `#ifdef`.  You don't even *have* to change this one right now; it's just a suggestion to make things nicer.

help/make_index.SH
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
```
dwfreed, line 2:
```suggestion

set -eu

```

include/serno.h.SH
```diff
@@ -0,0 +1,12 @@
+if [ -d .git ]; then
```
dwfreed, line 5:
```suggestion
#!/bin/sh

set -eu

if [ -d .git ]; then
```

ircd/version.c.SH
```diff
@@ -1,13 +1,10 @@
 #!/bin/sh
```
dwfreed, line 2:
```suggestion

set -eu

```

librb/src/version.c.SH
```diff
@@ -1,76 +1,24 @@
 #!/bin/sh
```
dwfreed, line 2:
```suggestion

set -eu

```

help/make_index.SH
dwfreed: > Doubling up the forward slashes is fine (`$DESTDIR/$2` correctly handles the case where $2 starts with a slash).

It doesn't correctly handle when `DESTDIR` is not defined, or is explicitly `/`, because then you can end up with double leading slash, which POSIX has explicitly left undefined.  I've personally worked in an environment that used this undefined behavior to its own advantage.

> The component `./` is a no-op.

By itself, sure.  Combined with `DESTDIR` with or without a `/` between, it absolutely is not.  The current behavior with a `/` between:

| `DESTDIR` | Location  |
|-----------|-----------|
| undefined | `/`       |
| defined   | `DESTDIR` |

This script is supposed to write indexes in either the `help` source directory, or in the `helpdir` installation directory, the latter of which needs to take `DESTDIR` into account.  The above is none of those.

Part of the problem is trying to keep the script working for 2 different use cases.  Eliminate one use case, and things become substantially similar.  The path forward would be to remove `index` as a dependency of `all`, leaving it only a dependency of `install`.  Then the second parameter to the script should always be the `helpdir` installation directory.  From there, the script can strip any trailing slash from `DESTDIR` (in case it is `/`) and any leading slash from the second parameter (in case `DESTDIR` is undefined) and combine them with a slash to get the right effect:

```sh
: ${DESTDIR:=/} # needed because of set -u just added
OUTPUT="${DESTDIR%/}/${2#/}"
```

dwfreed: Forgot to mention you should also ditch the help backup in `install`, because the `index` target will overwrite the indexes before the backup.

help/Makefile.am
dwfreed: This is the help backup I was talking about; this should go, because now the indexes have been modified already before this runs.  People should have their own backups.

help/make_index.SH
```diff
@@ -0,0 +1,45 @@
...
+USERS_INDEX="$OUTPUT/users/index"
+OPERS_INDEX="$OUTPUT/opers/index"
+
+# Build users index
+mkdir -p "$OUTPUT/users/"
+rm -f "$USERS_INDEX.tmp"
+for help in "$SOURCE/users/"*; do
+  if [ -f "$help" ]; then
+    echo $(basename "$help") >> "$USERS_INDEX.tmp"
```
internet-catte:
```suggestion
    basename "$help" >> "$USERS_INDEX.tmp"
```
this echo is useless

help/make_index.SH
```diff
@@ -0,0 +1,45 @@
...
+echo '' >> "$USERS_INDEX"
+cat "$USERS_INDEX.tmp" | sed -e 's|^users/||' | sort -u | tr a-z A-Z | column -c 65 -x | expand >> "$USERS_INDEX"
+rm -f "$USERS_INDEX.tmp"
+
+# Build opers index
+mkdir -p "$OUTPUT/opers/"
+rm -f "$OPERS_INDEX.tmp"
+for help in "$SOURCE/opers/"*; do
+  if [ -f "$help" ]; then
+    echo "$(basename $help)" >> "$OPERS_INDEX.tmp"
```
internet-catte:
```suggestion
    basename $help >> "$OPERS_INDEX.tmp"
```

help/make_index.SH
```diff
@@ -0,0 +1,45 @@
...
+rm -f "$USERS_INDEX.tmp"
+for help in "$SOURCE/users/"*; do
+  if [ -f "$help" ]; then
+    echo $(basename "$help") >> "$USERS_INDEX.tmp"
+  fi
+done
+
+# Add symlinks to users index
+for link in topic accept cmode admin names links away whowas version kick who invite quit join list nick oper part time credits motd userhost users whois ison lusers user help pass error challenge knock ping pong map trace chantrace extban monitor batch; do
+  echo $(basename "$link") >> "$USERS_INDEX.tmp"
```
internet-catte:
```suggestion
  basename "$link" >> "$USERS_INDEX.tmp"
```

dwfreed:
```suggestion
  echo "$link" >> "$USERS_INDEX.tmp"
```

Since the variable is already the basename

meson.build
```diff
@@ -0,0 +1,285 @@
...
+
+if mbedtls_lib.found() and mbedx509_lib.found() and mbedcrypto_lib.found()
+  tls_backend = 'mbedtls'
+  tls_deps = [mbedtls_lib, mbedx509_lib, mbedcrypto_lib]
+  have_mbedtls = true
+elif openssl_dep.found()
+  tls_backend = 'openssl'
+  tls_deps = [openssl_dep]
+  have_openssl = true
+elif tls_backend == 'none' and gnutls_dep.found()
```
aaronmdjones: This seems redundant (line 30 already ensures it is so)

```diff
-elif tls_backend == 'none' and gnutls_dep.found()
+elif gnutls_dep.found()
```

dwfreed: Fixed in the rebase

PR #496, opened 2026-02-12
Make TAGMSG behave like NOTICE for errors

skizzerz:
We explicitly do not send error numerics to clients for NOTICE in most cases, instead silently discarding the command. Give this treatment to TAGMSG across the board as well. Previously, this was the case for TAGMSG without tags but was not applied for TAGMSG that had tags but otherwise had issues (e.g. +m channel and the user is not +ov). This prevents spamming users whose clients send automatic TAGMSG with errors they can do nothing about.

### DIFF
--- a/modules/core/m_message.c
+++ b/modules/core/m_message.c
@@ -386,7 +386,7 @@ build_target_list(enum message_type msgtype, struct Client *client_p,
 			}
 
 			/* non existant channel */
-			else if(msgtype != MESSAGE_TYPE_NOTICE)
+			else if(msgtype == MESSAGE_TYPE_PRIVMSG)
 				sendto_one_numeric(source_p, ERR_NOSUCHNICK,
 						   form_str(ERR_NOSUCHNICK), nick);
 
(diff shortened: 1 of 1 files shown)


progval:
This behavior should probably depend on the tags.

For example, silently dropping `+react`/`+unreact` (once Solanum enables it) would be bad.

modules/core/m_message.c
```diff
@@ -887,7 +887,7 @@ msg_client(enum message_type msgtype,
 		 * would allow people to start filling up random users
 		 * targets just by ctcping them
 		 */
-		if((msgtype != MESSAGE_TYPE_NOTICE || *text != '\001') &&
+		if((msgtype == MESSAGE_TYPE_PRIVMSG || *text != '\001') &&
```
aaronmdjones, line 890: This change would allow TAGMSG to bypass tgchange throttling. Is that intentional?

skizzerz: Yes it's intentional, but more for the reasoning that "non-CTCP containing NOTICE doesn't and this change makes TAGMSG behave exactly like NOTICE" rather than any deeper logic. I'm honestly not sure why non-CTCP NOTICE is exempt from tgchange either though, and am of the personal opinion that nothing should really be exempt from it.

aaronmdjones: Scratch this, I just remembered that TAGMSG can't have text content.

aaronmdjones (approved):

Issue #484, opened 2026-01-21
m_part calls `h_privmsg_channel` hook without setting `msgbuf`


progval:
Since https://github.com/solanum-ircd/solanum/commit/0aac79db9d8747facb0fd59b673a437fa6f5696c, `hook_data_privmsg_channel` has a `msgbuf` member. But it is uninitialized here:

https://github.com/solanum-ircd/solanum/blob/86596936c1c4a708cb3ec3a642c008489d67255e/modules/core/m_part.c#L192-L200

which causes this error when a user sends a `PART`:

```
==4141408== Use of uninitialised value of size 8
==4141408==    at 0x4889AA8: msgbuf_get_tag (msgbuf.c:480)
==4141408==    by 0x48EC3F9: add_message_id_channel (tag_message_id.c:87)
==4141408==    by 0x487FC18: call_hook (hook.c:259)
==4141408==    by 0x593E584: do_message_hook (m_part.c:200)
==4141408==    by 0x593E584: part_one_client (m_part.c:130)
==4141408==    by 0x593E584: m_part (m_part.c:88)
==4141408==    by 0x48901D4: handle_command (parse.c:315)
==4141408==    by 0x48901D4: parse (parse.c:224)
==4141408==    by 0x488F66B: parse_client_queued.part.0 (packet.c:150)
==4141408==    by 0x488F8C4: parse_client_queued (packet.c:49)
==4141408==    by 0x488F8C4: read_packet (packet.c:306)
==4141408==    by 0x4B3E47B: rb_select_epoll (epoll.c:199)
==4141408==    by 0x4B3916F: rb_select (commio.c:2038)
==4141408==    by 0x4B3C37A: rb_lib_loop (rb_lib.c:219)
==4141408==    by 0x48820A0: solanum_main (ircd.c:756)
==4141408==    by 0x4950CA7: (below main) (libc_start_call_main.h:58)
==4141408==  Uninitialised value was created by a stack allocation
==4141408==    at 0x593E216: m_part (m_part.c:70)
==4141408== 
==4141408== Invalid read of size 8
==4141408==    at 0x4889AA8: msgbuf_get_tag (msgbuf.c:480)
==4141408==    by 0x48EC3F9: add_message_id_channel (tag_message_id.c:87)
==4141408==    by 0x487FC18: call_hook (hook.c:259)
==4141408==    by 0x593E584: do_message_hook (m_part.c:200)
[...]

Issue #396, opened 2023-01-11
null dereference in configure_authd


progval:
Happens since fbc97166a6e455f5cccf173abca3b9baedee4066 (cc @aaronmdjones):

```
$ cat /tmp/conf.txt
serverinfo {
    name = "My.Little.Server";
    sid = "42X";
    description = "test server";

};

general {
    throttle_count = 100;  # We need to connect lots of clients quickly
    # disable throttling for LIST and similar:
    pace_wait_simple = 0 second;
    pace_wait = 0 second;
    sasl_service = "SaslServ";
};

class "server" {
    ping_time = 5 minutes;
    connectfreq = 5 minutes;
};

listen {
    defer_accept = yes;

    host = "0.0.0.0";
    port = 33699;
    port = 57313;
};

auth {
    user = "*";
    flags = exceed_limit;
    
};

channel {
    disable_local_channels = no;
    no_create_on_split = no;
    no_join_on_split = no;
    displayed_usercount = 0;
};

connect "services.example.org" {
    host = "localhost";  # Used to validate incoming connection
    port = 0;  # We don't need servers to connect to services
    send_password = "password";
    accept_password = "password";
    class = "server";
    flags = topicburst;
};
service {
    name = "services.example.org";
};

privset "omnioper" {
    privs = oper:general, oper:privs, oper:testline, oper:kill, oper:operwall, oper:message,
            oper:routing, oper:kline, oper:unkline, oper:xline,
            oper:resv, oper:cmodes, oper:mass_notice, oper:wallops,
            oper:remoteban,
            usermode:servnotice, auspex:oper, auspex:hostname, auspex:umodes, auspex:cmodes,
            oper:admin, oper:die, oper:rehash, oper:spy, oper:grant;
};
operator "operuser" {
    user = "*@*";
    password = "operpassword";
    privset = "omnioper";
    flags = ~encrypted;
};

$ valgrind solanum -foreground -configfile /tmp/conf.txt
==3377160== Memcheck, a memory error detector
==3377160== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3377160== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==3377160== Command: solanum -foreground -configfile /tmp/conf.txt
==3377160== 
  notice: starting solanum-1.0-dev ...
[...]

PR #497, opened 2026-02-14
Run irctest in CI

progval:
Based on https://github.com/inspircd/inspircd/blob/148a90535c1b061efcc13439056f8ab786086084/.github/workflows/ci-irctest.yml

It caught these regressions in the past:

* https://github.com/solanum-ircd/solanum/issues/396
* https://github.com/solanum-ircd/solanum/pull/462
* https://github.com/solanum-ircd/solanum/pull/463
* https://github.com/solanum-ircd/solanum/issues/484
* https://github.com/solanum-ircd/solanum/pull/535


.github/workflows/ci.yml
```diff
@@ -87,3 +87,116 @@ jobs:
...
+      run: |-
+        cd ~
+        tar -czf artifacts-solanum.tar.gz ircd/
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v6
+      if: matrix.os == 'ubuntu-24.04' && matrix.compiler == 'gcc-13'
+      with:
+        name: installed-solanum-${{matrix.os}}-${{matrix.compiler}}
+        path: ~/artifacts-solanum.tar.gz
+        retention-days: 1
```
progval, line 156: I chose to make one of the existing builds (the fastest one) export its artifacts at the end for the irctest job. This minimizes CPU usage, but unnecessarily increases the wall-clock run time because the dependency means the `test` job waits for every `build` job to be done.

I can make the `test` jobs have its their own build concurrently to the other ones if you prefer.

internet-catte (approved):

PR #507, opened 2026-03-18
Add +reply client tag support

skizzerz:
We validate that the reply looks like a valid message id as generated by tag_message_id and that the reply is going to the same target that could have generated the message id. Since tag values do not currently go through spamfilter, this is an easy-ish way to ensure that it cannot be used as a backchannel for spam.

### DIFF
--- a/extensions/tag_reply.c
+++ b/extensions/tag_reply.c
@@ -1,0 +1,143 @@
+/*
+ * Solanum: a slightly advanced ircd
+ * tag_reply.c: implement the IRCv3 +reply client tag
+ *
+ * Copyright (c) 2026 Ryan Schmidt <skizzerz@skizzerz.net>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "stdinc.h"
+#include "modules.h"
+#include "hook.h"
+#include "client.h"
+#include "client_tags.h"
+#include "ircd.h"
+#include "send.h"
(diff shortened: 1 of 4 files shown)


JustAnotherArchivist:
> I would be fairly surprised (as an end user) if I clicked on a reply breadcrumb only to [...] be told the original message doesn't exist because it's from a channel or PM I'm not on.

This is a case you have to handle as a client anyway though even with your same-target restriction: it could be a reply to an older message from when you weren't yet in the channel (and which isn't included in chathistory, if applicable in the future) or a reply to a message that's only delivered _later_, e.g. on lag between servers.

For cross-target replies, sure, simply jumping to another channel would be poor UX (at least by default), but that's not the only UI option. Personally, I'd probably just prompt the user with 'this is a reply to a message in <target>, do you want to go there?'-type message. Either way, unless there's an agreement among ircd devs to not support cross-target replies (in which case it should probably go into the spec), again, it's something the clients will need to consider anyway in the implementation design.

extensions/tag_reply.c
```diff
@@ -0,0 +1,148 @@
...
+			|| strchr(data->message->para[1], ',') != NULL)
+		{
+			return;
+		}
+
+		/* check if the target is a channel (possibly a statusmsg) */
+		const char *ch_target = NULL;
+		if (IsChannelName(data->message->para[1]))
+			ch_target = data->message->para[1];
+		else if (IsChannelName(data->message->para[1] + 1))
```
aaronmdjones: This should also check if the first character is a channel status prefix IMO

extensions/tag_reply.c
```diff
@@ -0,0 +1,148 @@
...
+		if ((ch_target == NULL) ^ (idlen == MSGID_LEN_MIN))
+			return;
+
+		/* quick validation of msgid portion before channel name */
+		for (int i = 1; i < 29; i++)
+		{
+			if (i < 20 && !isdigit(data->value[i]))
+				return;
+			if (i >= 20 && !isupper(data->value[i]) && !isdigit(data->value[i]))
+				return;
```
aaronmdjones: This can be simplified to

```C
if (isdigit(data->value[i]))
    continue;
if (i >= 20 && isupper(data->value[i]))
    continue;
return;
```

extensions/tag_reply.c
```diff
@@ -0,0 +1,148 @@
...
+
+			bool is_match = !irccmp(chname, ch_target);
+			rb_free(chname);
+
+			if (!is_match)
+				return;
+
+			struct Channel *chptr = find_channel(ch_target);
+			if (chptr == NULL)
```
aaronmdjones: You're not using `chptr` here, just simplify it to

```C
if (find_channel(ch_target) == NULL)
```

aaronmdjones (approved):

PR #511, opened 2026-03-23
Implementation of labeled-response

skizzerz:
This was broken into multiple commits for easier review as there are a number of distinct things changed.

- Expand outbound_msgbuf hook to use a different hook data. First three fields remain compatible with the original hook_data struct for a semblance of ABI compatibility. (that is, newer core with older modules. older core with new modules that use new fields will crash)
- new parse_end hook for automatic sending of labelled ACK and closing labeled-response batches
- STATS t expanded to show details about pending client-initiated and labeled-response batches
- new servonly clicap for propagation of s2s tags that must never get sent to actual clients
- new clicap flag so message tags with it as its capmask will not propagate between servers
- ENCAP ACK s2s command to signal completion of a remote batch, automatically sent by parse.c
- SAFELIST and BATCH needed to gain tracking for labeled-response
- Removal of opers@server special message target

### DIFF
--- a/ircd/batch.c
+++ b/ircd/batch.c
@@ -22,6 +22,7 @@
 #include "stdinc.h"
 #include "batch.h"
 #include "rb_dictionary.h"
+#include "response.h"
 
 static rb_dictionary *handlers = NULL;
 
@@ -65,9 +66,6 @@ generate_batch_id(char *buf, size_t size)
 {
 	static const char *alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-A";
 
-	if (size > BATCH_ID_LEN + 1)
-		size = BATCH_ID_LEN + 1;
-
 	for (size_t i = 0; i < size - 1; i++)
 		buf[i] = alphabet[rand() % 64];
 
@@ -109,6 +107,7 @@ batch_free(struct Batch *batch)
 		batch_free(child);
 	}
 
+	free_response_batch(batch->response_info);
 	rb_free(batch->start->data);
 	rb_free(batch->start);
 	rb_free(batch);
--- a/ircd/batch.c
+++ b/ircd/batch.c
@@ -122,8 +122,12 @@ allocate_batch_message(struct MsgBuf *msg)
 
 	for (int i = 0; i < msg->n_tags; i++)
 	{
+		if (msg->tags[i].key == NULL)
+			continue;
+
 		len += strlen(msg->tags[i].key) + 1;
-		len += strlen(msg->tags[i].value) + 1;
+		if (msg->tags[i].value != NULL)
+			len += strlen(msg->tags[i].value) + 1;
 	}
 
 	for (int i = 0; i < msg->n_para; i++)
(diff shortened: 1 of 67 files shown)


ircd/response.c
```diff
@@ -0,0 +1,185 @@
...
+void
+begin_remote_response_batch(int server_count)
+{
+	if (outgoing_response_info == NULL || !MyConnect(outgoing_response_info->source_p))
+		return;
+
+	if (!EmptyString(outgoing_response_info->batch))
+	{
+		/* increase the number of servers we're expecting responses from */
+		outgoing_response_info->remote_response += server_count;
```
aaronmdjones: This will be overwritten on line 115

ircd/response.c
```diff
@@ -0,0 +1,185 @@
...
+
+		/* if we're upgrading a local batch to a remote batch, add it to tracking places */
+		if (outgoing_response_info->expires == 0)
+		{
+			outgoing_response_info->expires = rb_current_time() + RESPONSE_EXPIRY;
+			rb_dlinkAddAlloc(outgoing_response_info, &outgoing_response_info->source_p->localClient->pending_remote_responses);
+			rb_dictionary_add(pending_responses, outgoing_response_info->batch, outgoing_response_info);
```
aaronmdjones: These will be repeated on lines 117-118

ircd/response.c
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
```
aaronmdjones, line 124: Have you forgotten a return for the above reasons?

ircd/batch.c
```diff
@@ -146,15 +150,28 @@ allocate_batch_message(struct MsgBuf *msg)
 
 	copy->msg.n_tags = msg->n_tags;
 	copy->msg.tagslen = msg->tagslen;
+	int j = 0;
```
aaronmdjones: There's no need for this variable to live outside of the loop

```diff
-    int j = 0;
-    for (int i = 0; i < msg->n_tags; i++)
+    for (int i = 0, j = 0; i < msg->n_tags; i++)
```

modules/cap_labeled_response.c
```diff
@@ -0,0 +1,230 @@
...
+}
+
+static void
+me_ack(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source_p, int parc, const char *parv[])
+{
+	/* client already left or something, so nothing to do here */
+	if (outgoing_response_info == NULL)
+		return;
+
+	outgoing_response_info->remote_response--;
```
aaronmdjones, line 237: There is no protection here for underflow. What happens if a server is introduced remotely after this is initially set, and starts replying with ACKs this server wasn't expecting?

ircd/response.c
skizzerz: I did indeed forget a return! Although I think I'll wrap the bottom part into an else block instead to make the intent clearer.

modules/cap_labeled_response.c
skizzerz: Can't happen. Once the count hits 0, we call `free_response_batch` on line 208 which deletes our tracking for that batch ID. When the additional unexpected ACKs come in, `get_remote_response_batch` on line 116 will return NULL, therefore `outgoing_response_info` will be NULL. Because `outgoing_response_info` is NULL, the test on line 224 will fail and we will not decrement anything further (or dereference NULL).

ircd/response.c
aaronmdjones: Aha. Well, if you're going to wrap it in an else, you may as well swap the blocks around to avoid a pointless negation in if. :)

modules/cap_labeled_response.c
aaronmdjones: Fair enough.

skizzerz: I adjusted the cleanup event in response.c to free these batches regardless even though it shouldn't be possible to get there, just as a failsafe.

aaronmdjones (dismissed):

aaronmdjones (dismissed):

aaronmdjones (approved):

PR #513, opened 2026-03-26
Allow all ircd file paths to be customized

skizzerz:
The ban.db file as well as both MOTD files were in hardcoded locations. Add new config options to the main solanum binary to manipulate where these files are looked for. This should make it fully possible to run multiple solanum instances from the same directory side-by-side.


ircd/ircd.c
```diff
@@ -261,6 +264,12 @@ struct lgetopt myopts[] = {
 	 STRING, "File to use for ircd.log"},
 	{"pidfile", &pidFileName,
 	 STRING, "File to use for process ID"},
+	{"banfile", &banFileName,
+	 STRING, "File to use for ban.db"},
+	{"motdfile", &banFileName,
```
aaronmdjones: Copy/paste error

aaronmdjones (changes requested):

aaronmdjones (approved):

PR #518, opened 2026-04-12
Implement bot mode

### LINKED ISSUE #210: Support umode +B for bots

TheDaemoness:
Needs testing before merge.

Closes #210
Closes #278


extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+	hook_data_umode_changed *data = data_;
+
+	/* Check if we are in at least one channel. */
+	if (data->client->user->channel.head == NULL)
+		return;
+
+	if (IsBot(data->client) && !(data->oldumodes & umode))
+	{
+		/* Attempted to set botmode while in a channel. TODO: Send error. */
+		data->client->umodes &= !umode;
```
dwfreed:
```suggestion
		data->client->umodes &= ~umode;
```

This needs to be bitwise invert, not boolean

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+moddeinit(void)
+{
+	user_modes['B'] = 0;
+	construct_umodebuf();
+	delete_isupport("BOT");
+	if (chmode)
+		cflag_orphan('B');
+}
+
+void
```
dwfreed:
```suggestion
static void
```

All functions in modules should be `static`.

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+		return;
+
+	/* If we're here, umode AND cmode +B are set, and the client is not exempt for any reason. */
+	sendto_one_numeric(data->client, ERR_CANNOTSENDTOCHAN,
+			"%s :Cannot send to channel (+B) - bots must be opped/voiced",
+			data->chptr->chname);
+	data->approved = CAN_SEND_NO;
+}
+
+void
```
dwfreed:
```suggestion
static void
```

Likewise

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	/* If we're here, umode AND cmode +B are set, and the client is not exempt for any reason. */
+	sendto_one_numeric(data->client, ERR_INVITEONLYCHAN,
+			"%s :Cannot join channel (+B) - bots must be invited",
+			chptr->chname);
+	data->approved = ERR_CUSTOM;
+}
+
+void
```
dwfreed:
```suggestion
static void
```

Likewise

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+botmode_apply_tag(void *data_)
+{
+	hook_data *data = data_;
+	struct MsgBuf *msgbuf = data->arg1;
+
+	if (data->client != NULL && IsBot(data->client) && *data->client->user->suser)
+		msgbuf_append_tag(msgbuf, "bot", data->client->user->suser, CLICAP_MESSAGE_TAGS);
+}
+
+void
```
dwfreed:
```suggestion
static void
```

Likewise

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+botmode_whois(void *data_)
+{
+	hook_data_client *data = data_;
+	if(!IsBot(data->target))
+		return;
+
+	sendto_one_numeric(data->client, RPL_WHOISBOT, form_str(RPL_WHOISBOT), data->client->name, data->target->name);
+}
+
+void
```
dwfreed:
```suggestion
static void
```

Likewise

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+
+static void
+moddeinit(void)
+{
+	user_modes['B'] = 0;
+	construct_umodebuf();
+	delete_isupport("BOT");
+	if (chmode)
+		cflag_orphan('B');
+}
```
dwfreed:
```suggestion
	chmode = cflag_add('B', chm_simple);
	if (!chmode)
	{
		user_modes['B'] = 0;
		return -1;
	}
	
	construct_umodebuf();
	add_isupport("BOT", isupport_string, "B");
	return 0;
}

static void
moddeinit(void)
{
	user_modes['B'] = 0;
	construct_umodebuf();
	delete_isupport("BOT");
	cflag_orphan('B');
}
```

Need to check for `cflag_add` failure too, or you end up with silent behavior differences.

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+
+void
+botmode_apply_tag(void *data_)
+{
+	hook_data *data = data_;
+	struct MsgBuf *msgbuf = data->arg1;
+
+	if (data->client != NULL && IsBot(data->client) && *data->client->user->suser)
+		msgbuf_append_tag(msgbuf, "bot", data->client->user->suser, CLICAP_MESSAGE_TAGS);
```
dwfreed:
```suggestion
		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
```

"Servers MUST NOT send this tag with a value"

dwfreed (changes requested):
I make no comment at the present time on the usefulness of this, but that should be discussed further before this is merged.

extensions/botmode.c
```diff
@@ -0,0 +1,203 @@
...
+		data->client->umodes &= !umode;
+	}
+	else if (!IsBot(data->client) && (data->oldumodes & umode))
+	{
+		/* Attempted to unset botmode while in a channel. TODO: Send error. */
+		data->client->umodes |= umode;
+	}
+}
+
+mapi_hfn_list_av1 botmode_hfnlist[] = {
```
dwfreed:
```suggestion
static mapi_hfn_list_av1 botmode_hfnlist[] = {
```

This also should be static

extensions/botmode.c
```diff
@@ -0,0 +1,207 @@
...
+#include "send.h"
+#include "s_conf.h"
+#include "s_serv.h"
+#include "s_user.h"
+#include "supported.h"
+
+static const char botmode_desc[] =
+	"Implements bot mode and adds a channel mode to prohibit bots from joining or speaking";
+
+static unsigned chmode, umode;
```
skizzerz:
```suggestion
static unsigned int chmode, umode;
```

If I (or someone else) ever needs to move these to be 64-bit in the future, having something consistent to be able to grep for is nice. Omitting "int" in "unsigned int" is imo a massive misfeature of C and caused some bugs in my conversion of the cap bitfields.

extensions/botmode.c
```diff
@@ -0,0 +1,207 @@
...
+
+#define IsBot(client) ((client)->umodes & umode)
+
+static int
+modinit(void)
+{
+	umode = user_modes['B'] = find_umode_slot();
```
skizzerz: check if `user_modes['B']` is nonzero and abort before overwriting it? (prevents clashes with 3rd party modules that are already using the mode)

extensions/botmode.c
```diff
@@ -0,0 +1,207 @@
...
+		return;
+
+	if (!IsBot(data->client))
+		return;
+
+	if (!(data->chptr->mode.mode & chmode))
+		return;
+
+	/* Allow oper bots to bypass this mode. */
+	if (IsOper(data->client))
```
skizzerz: Make this a new priv, or re-use oper:message/oper:always_message? I don't see why oper bots would need automatic bypass regardless of privset.

extensions/botmode.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+static void
+botmode_whois(void *data_)
+{
+	hook_data_client *data = data_;
+	if(!IsBot(data->target))
+		return;
+
+	sendto_one_numeric(data->client, RPL_WHOISBOT, form_str(RPL_WHOISBOT), data->client->name, data->target->name);
```
progval:
```suggestion
	sendto_one_numeric(data->client, RPL_WHOISBOT, form_str(RPL_WHOISBOT), data->target->name);
```

`sendto_one_numeric` already includes the sender's nick

include/messages.h
```diff
@@ -107,6 +107,7 @@
 #define NUMERIC_STR_331      ":%s 331 %s %s :No topic is set."
 #define NUMERIC_STR_332      ":%s 332 %s %s :%s"
 #define NUMERIC_STR_333      ":%s 333 %s %s %s %lld"
+#define NUMERIC_STR_335      "%s %s :is a bot"
```
progval:
```suggestion
#define NUMERIC_STR_335      "%s :is a bot"
```

see above

progval (commented):
The spec also [defines a message tag](https://ircv3.net/specs/extensions/bot-mode#the-bot-tag):

> The bot tag indicates that the given user is a bot. This tag SHOULD be added by the server to all commands sent by a bot (e.g. PRIVMSG, JOIN, MODE, NOTICE, and all others). The tag SHOULD also be added by the ircd to all numerics directly caused by the bot. This tag MUST only be sent to users who have requested the message-tags capability. Servers MUST NOT send this tag with a value, and clients MUST ignore any value if it exists.

Will it be implemented in a future PR?

TheDaemoness:
> Will [the bot tag] be implemented in a future PR?

The first "SHOULD" clause (attaching the tag to messages sent by the bot) should be entirely covered by `botmode_apply_tag` which gets called by the `outbound_msgbuf` hook, assuming my implementation isn't currently broken (which it very well could be, this PR is a draft for a reason :grimacing:).
I'm not sure about the second clause (attaching the tag to numerics directly caused by the bot). I think it would depend on whether the bot is considered the source of those numerics or not.

extensions/botmode.c
```diff
@@ -0,0 +1,211 @@
...
+}
+
+static void
+botmode_apply_tag(void *data_)
+{
+	hook_data *data = data_;
+	struct MsgBuf *msgbuf = data->arg1;
+
+	if (data->client != NULL && IsBot(data->client) && *data->client->user->suser)
+		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
```
progval:
```suggestion
	if (data->client != NULL && IsBot(data->client))
		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
```

otherwise the tag is not sent for logged out bots:

```
1777791985.960 user -> S: WHOIS botnick
1777791985.960 S -> user: :My.Little.Server 311 usernick botnick ~username 127.0.0.1 * :Realname
1777791986.001 S -> user: :My.Little.Server 312 usernick botnick My.Little.Server :test server
1777791986.001 S -> user: :My.Little.Server 338 usernick botnick 127.0.0.1 :actually using host
1777791986.001 S -> user: :My.Little.Server 317 usernick botnick 0 1777791985 :seconds idle, signon time
1777791986.001 S -> user: :My.Little.Server 335 usernick botnick :is a bot
1777791986.001 S -> user: :My.Little.Server 318 usernick botnick :End of /WHOIS list.
1777791986.001 S -> user: :My.Little.Server PONG My.Little.Server :synchronize7434.098827309

1777791986.002 bot -> S: PRIVMSG usernick :beep boop
1777791986.002 S -> user: @msgid=1177779198600101577642XAAAAAC :botnick!~username@127.0.0.1 PRIVMSG usernick :beep boop
```

(copy-pasted from the account-tag module?)

extensions/botmode.c
```diff
@@ -0,0 +1,211 @@
...
+{
+	hook_data_umode_changed *data = data_;
+
+	/* Check if we are in at least one channel. */
+	if (data->client->user->channel.head == NULL)
+		return;
+
+	if (IsBot(data->client) && !(data->oldumodes & umode))
+	{
+		/* Attempted to set botmode while in a channel. TODO: Send error. */
```
internet-catte: did we decide whether to send an error for this (and in the `else if`)?

internet-catte (approved):
This seems sound to my non-expert eye

extensions/botmode.c
TheDaemoness: I'm not sure we did. IIRC the options are as follows:

* Send nothing, as is done for some other non-oper umodes. As the simplest solution and one that we can reasonably change later, that's what's implemented here.
* Add numeric [673](https://defs.ircdocs.horse/defs/numerics#rpl-cannotsetmodes-673) and hope that some clients understand it.
* Implement the `standard-replies` cap, and for clients that request it, send a `FAIL MODE` message.

For the other two options, it would make sense to also implement them for the other user modes. Worth doing in a future PR, perhaps?

extensions/botmode.c
```diff
@@ -0,0 +1,211 @@
...
+
+	chmode = cflag_add('B', chm_simple);
+	if (!chmode)
+	{
+		user_modes['B'] = 0;
+		return -1;
+	}
+
+	construct_umodebuf();
+	add_isupport("BOT", isupport_string, "B");
```
Steering7253: Presumably this should be isupport_umode?

extensions/botmode.c
Steering7253: The client shouldn't assume the mode has been accepted until it sees the MODE command repeated back at it, anyway. Try setting `MODE yourself +o` and see what the ircd does.

extensions/botmode.c
```diff
@@ -0,0 +1,211 @@
...
+}
+
+static void
+botmode_apply_tag(void *data_)
+{
+	hook_data *data = data_;
+	struct MsgBuf *msgbuf = data->arg1;
+
+	if (data->client != NULL && IsBot(data->client))
+		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
```
skizzerz: Does the bot tag need to be propagated s2s? Right now it is as-written. umodes get propagated so each server will be able to perform this evaluation on its own. I recommend that you only apply this tag to messages sent to local clients (which in turn means that you'll need to rebase on PR 511 because the capability to do that wasn't introduced until that branch)

extensions/botmode.c
TheDaemoness: You're right, applying.

dwfreed: This isn't necessary because you can't put bot mode in `oper_only_umodes` anyway

extensions/botmode.c
TheDaemoness: Silent failures are a protocol design flaw IMO. Aside from the client not knowing for sure that a failure has occurred until the next unrelated message or until a timeout happens, the client isn't told why the failure occurred, leaving the user to either search for documentation or ask someone who might know. That is, assuming the user even notices that the failure occurred.

But I do think that fixing the status quo for failed mode changes is out-of-scope for this PR, and as you've noted, clients must be able to handle silent failure.

extensions/botmode.c
TheDaemoness: I agree, but would it be out of the question to switch to non-propagation after #511 gets merged? EDIT: It got merged.

extensions/botmode.c
Steering7253: Yeah - to clarify, IMO there should be some sort of error response for failed modes, but that's a larger issue and whether clients "understand" it (fully) isn't super important :) the error is primarily for the user operating the client

extensions/botmode.c
TheDaemoness: Should I abstain from making a commit that changes this, then?

EDIT: Discussed on IRC, abstaining from changing it at this time.

extensions/botmode.c
TheDaemoness: Per internal discussion with @skizzerz, this isn't actually feasible with the hook data as-is, because the `bot` tag is not gated by a particular capability that can have its tags made non-relayable, and `data->target` cannot be used to reliably check for the circumstance where the `MsgBuf` is being constructed in preparation for being sent to servers.

furudean:
What is the justification for not allowing changes to bot mode if the user is currently in a channel?

TheDaemoness:
There was some internal discussion among several members of Libera.Chat staff about this prior to me making the draft PR. This behavior is the consequence of four opinions (listed in no particular order):

1. Bot mode should ideally last for close to the lifetime of the connection, accepting that currently there's no decently-supported way to set user modes during connection registration (no, passing under-specified bitsets in `USER` does not count). 
2. The fact that the spec does not describe a mechanism by which clients may be proactively informed about a change in bot mode for clients in a shared channel is a mistake.
3. Behavior where users can mistakenly set the mode after joining channels and then be unable to unset it without disruption should be avoided. For this purpose, reconnecting should be assumed to be disruptive by default, considering cases like bouncers where it may not be self-evident to users how to terminate the unwanted session.
4. Users should be able to reasonably assume that clients with bot mode in a channel with cmode `B` are permitted by the channel's operators to be there.

If, after reading that, the current behavior seems like a compromise, then yes, yes it is,
and I'm open to suggestions that meet the above criteria.
There were a few other ideas considered:

- Additionally requiring the client to be in the flood grace period to change the mode, which would have the benefit of ensuring that PMs also correctly get the `bot` tag. Ruled out as it's too harsh to clients that send messages to NickServ (e.g. `IDENTIFY`) before setting modes. Arguably runs afoul of opinion 3, too.
- Having the client be forcibly parted from channels upon changing the mode ala `JOIN 0`. Ruled out due to being disruptive and something users could reasonably be tricked into doing (ala `JOIN 0`).
- Evicting clients from channels with cmode `B` if they set bot mode. This would be inconsistent with every other cmode with a join restriction.

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+	user_modes['B'] = 0;
+	construct_umodebuf();
+	delete_isupport("BOT");
+	cflag_orphan('B');
+}
+
+static void
+botmode_can_send(void *data_)
+{
+	hook_data_channel *data = data_;
```
skizzerz: The [can_send hook](https://github.com/solanum-ircd/solanum/blob/main/doc/technical/hooks.md#can_send) takes a hook_data_channel_approval, not hook_data_channel.

Because this hook fires on both local and remote clients, we probably also want a short-circuit on a remote client. The remote server already checked permissions, and operating on remote clients would cause duplicative checks at best and instances where some servers can see messages and others can't in the event of cmode desyncs.

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+		return;
+
+	if (!(data->chptr->mode.mode & chmode))
+		return;
+
+	/* Allow bots with oper:always_message to bypass this mode. */
+	if (HasPrivilege(data->client, "oper:always_message"))
+		return;
+
+	msptr = find_channel_membership(data->chptr, data->client);
```
skizzerz: This already exists as `data->msptr`

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+static void
+botmode_can_send(void *data_)
+{
+	hook_data_channel *data = data_;
+	struct membership *msptr;
+
+	/* If message is already blocked, defer. */
+	if (data->approved == CAN_SEND_NO)
```
skizzerz: Perhaps defer if `data->approved != CAN_SEND_NONOP`? This catches both the case where it's already rejected, as well as the case where it is `CAN_SEND_OPV` (indicating the user is opped/voiced in the channel, and thus letting you skip the `is_chanop_voiced` check further down)

extensions/chm_nobots.c
```diff
@@ -0,0 +1,212 @@
...
+		return;
+
+	if (!IsBot(client))
+		return;
+
+	if (!(chptr->mode.mode & chmode))
+		return;
+
+	/* Allow oper bots to bypass this mode. */
+	if (IsOper(data->client))
```
skizzerz, line 111: Do you want IsOperGeneral here instead (or a new priv)? IsOper just tests for umode +o and has no priv checks, so it'd be largely impossible to have an oper bot that *wants* to honor the cmode.

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+
+	if (data->source != NULL && IsBot(data->source))
+		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
+}
+
+static void
+botmode_whois(void *data_)
+{
+	hook_data_client *data = data_;
+	if(!IsBot(data->target))
```
skizzerz:
```suggestion
	if (!IsBot(data->target))
```

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+static mapi_hfn_list_av1 botmode_hfnlist[] = {
+	{ "can_send", botmode_can_send },
+	{ "can_join", botmode_can_join },
+	{ "outbound_msgbuf", botmode_apply_tag },
+	{ "doing_whois", botmode_whois },
+	{ "doing_whois_global", botmode_whois },
+	{ "umode_change", botmode_change },
```
skizzerz:
```suggestion
	{ "umode_changed", botmode_change },
```

extensions/botmode.c
```diff
@@ -0,0 +1,212 @@
...
+	sendto_one_numeric(data->client, RPL_WHOISBOT, form_str(RPL_WHOISBOT), data->target->name);
+}
+
+static void
+botmode_change(void *data_)
+{
+	hook_data_umode_changed *data = data_;
+
+	/* Check if we are in at least one channel. */
+	if (data->client->user->channel.head == NULL)
```
skizzerz:
```suggestion
	if (data->client->user->channel.length == 0)
```

Makes it a bit more clear IMO at a glance (or alternatively the `rb_dlink_list_length` macro but that's *more* typing instead of less so I really don't know why it exists).

modules/m_who.c
```diff
@@ -500,8 +501,11 @@ do_who(struct Client *source_p, struct Client *target_p, struct membership *mspt
 	size_t pos;
 	const char *q;
 
-	sprintf(status, "%c%s%s",
-		   target_p->user->away ? 'G' : 'H', SeesOper(target_p, source_p) ? "*" : "", msptr ? find_channel_status(msptr, fmt->fields || IsClientCapable(source_p, CLICAP_MULTI_PREFIX)) : "");
+	sprintf(status, "%c%s%s%s",
```
skizzerz: Since you're touching this anyway, switch to `snprintf`. The current code will never overflow the buffer but no reason not to use the safer function.

skizzerz (changes requested):

extensions/chm_nobots.c
TheDaemoness: I don't think `oper:general` needs more things rolled into it. I wouldn't be against adding a new priv, though. Name suggestions? `oper:bypass_chm_nobots`?

extensions/botmode.c
jillest: Since the check is cheap, I think it's fine to do it everywhere. This ensures there won't be a message apparently coming in after a mode change.

Note that cmode +m is checked everywhere, while bans and quiets are not, so there is no consistency here.

extensions/umode_bot.c
```diff
@@ -0,0 +1,100 @@
...
+	struct MsgBuf *msgbuf = data->msgbuf;
+
+	if (data->source != NULL && IsBot(data->source))
+		msgbuf_append_tag(msgbuf, "bot", NULL, CLICAP_MESSAGE_TAGS);
+}
+
+static void
+umode_bot_whois(void *data_)
+{
+	hook_data_client *data = data_;
```
skizzerz: This is now `hook_data_client_approval` (the rest of the hook function does not need any changes however)

extensions/chm_nobots.c
```diff
@@ -0,0 +1,150 @@
...
+
+	if (!(data->chptr->mode.mode & chmode))
+		return;
+
+	/* Allow bots with oper:always_message to bypass this mode. */
+	if (HasPrivilege(data->client, "oper:always_message"))
+		return;
+
+	/* If we're here, umode AND cmode +B are set, and the client is not exempt for any reason. */
+	sendto_one_numeric(data->client, ERR_CANNOTSENDTOCHAN,
```
skizzerz: The caller of can_send already sends ERR_CANNOTSENDTOCHAN if the hook fails, so this will result in duplicate messages. This line should be removed (the error message sent by the caller is not customizable, unfortunately, so it will just be a generic "cannot send" message).

can_send is *also* called repeatedly in contexts of checking sending permissions internally, when we are *not* actually sending a message. So, we definitely do not want this giving any output inside of the hook function.

extensions/chm_nobots.c
skizzerz: IsOper seems fine enough for now; we can re-evaluate later if we ever run into an oper bot that needs to care about the cmode for whatever reason.

extensions/restrict_botmode.c
```diff
@@ -0,0 +1,73 @@
...
+	unsigned int umode = user_modes['B'];
+	/* No need to check if umode is 0; if it is for some reason, both macros will evaluate to false. */
+
+	/* Check if we are in at least one channel. */
+	if (data->client->user->channel.length == 0)
+		return;
+
+	if (HasMode(data, umode) && !HadMode(data, umode))
+	{
+		/* TODO: Send error. */
```
skizzerz: For these TODOs, error should only be sent if `MyClient(data->client)` since this is called for both local and remote clients (we do want the logic to execute for remote clients since the MODE has already propagated, just no error messages). Best error mechanism right now is likely `sendto_one_notice` to send a server notice.

skizzerz (changes requested):

aaronmdjones (approved):

aaronmdjones:
GitHub will not let me merge this as a merge commit or as a rebase commit. I've rebased it locally and will force push it over.

PR #520, opened 2026-05-04
Clean up include/stdinc.h and librb/include/rb_lib.h

aaronmdjones:
- Unconditionally require `stdbool.h` in IRCd

  We are a C99 project anyway (`configure.ac` and `meson.build` both require a C99 compiler), so this header will always be present

  This removes the awful hack defining `bool`, `true`, and `false` in its absense, which was creating build errors when using meson, because `meson.build` did not set the `HAVE_STDBOOL_H` macro

- Modernise definition of `alloca(3)` according to GNU guidance

  This was in both librb and IRCd, while it only needs to be in the former

- Fix missing endianness check in `librb/meson.build`

  IRCd and librb both use the `WORDS_BIGENDIAN` macro, which was only being set by their `configure.ac` (with the `AC_C_BIGENDIAN` M4 macro)

  Since IRCd's `stdinc.h` includes librb's `rb_lib.h`, which then includes librb's `librb-config.h`, which would have the macro defined there, we only need to check for it in librb, so we can remove the endianness check from IRCd's `configure.ac` and don't need to add it to IRCd's `meson.build`

- Use `__has_attribute` to detect whether compiler attributes are available

  This is supported by gcc (5+), clang (2.9+), and ICC (17+)

- Add a `__noreturn` macro for `__attribute__((__noreturn__))` and adjust the codebase to use that

- Remove inclusion of headers that are only used by librb

- Remove inclusion of headers that are included in `rb_lib.h` because `stdinc.h` includes `rb_lib.h` anyway

- Remove the checks for headers and functions from the IRCd `configure.ac` and `meson.build` that are only used by librb

- Alphabetise checks for headers and functions (by the name of the macro they define) in the `meson.build` files

- Use `check_header()` instead of `has_header()` in meson to see if it is actually possible to compile code using the header

- Remove dead code using `srand48(3)` which has never been compiled because neither build system has ever set the `HAVE_SRAND48` macro and the surrounding code makes no use of this libc RNG

- Remove the TLS defines from IRCd's `meson.build` because librb's `meson.build` already defines them and IRCd pulls those in

### DIFF
--- a/authd/authd.c
+++ b/authd/authd.c
@@ -25,7 +25,7 @@
 
 #define MAXPARA 10
 
-static void error_cb(rb_helper *helper) __attribute__((noreturn));
+static void error_cb(rb_helper *helper) __noreturn;
 static void handle_reload(int parc, char *parv[]);
 static void handle_stat(int parc, char *parv[]);
 static void handle_options(int parc, char *parv[]);
--- a/authd/res.c
+++ b/authd/res.c
@@ -224,9 +224,6 @@ static void start_resolver(void)
  */
 void init_resolver(void)
 {
-#ifdef HAVE_SRAND48
-	srand48(rb_current_time());
-#endif
 	start_resolver();
 }
 
--- a/bandb/bandb.c
+++ b/bandb/bandb.c
@@ -237,7 +237,7 @@ parse_request(rb_helper *helper)
 
 
 static void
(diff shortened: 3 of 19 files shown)


skizzerz:
stdinc.h is missing include guards, should we add them in? Otherwise it's getting included multiple times in some translation units.

dwfreed (approved):

PR #522, opened 2026-05-13
Add solanum.chat/echo s2s batch type

skizzerz:
This batch is gated behind a new server capab ECHOB. It allows for echoing multiple lines of text within a batch (potentially with nested batches) for the echo-message mechanism. Unlike most s2s messages, any contents within a solanum.chat/echo batch are sent to the client exactly as-is; they are *not* parsed to e.g. replace UID/SIDs with the relevant client-facing identifiers. Limited parsing to validate the origin/prefix is not coming from a fake direction is still performed, so inner messages cannot have spoofed senders. Instead, the sender information is encoded in batch parameters so the destination server can spoof that information when delivering to the client.

While nothing in the codebase currently uses this, it is used for the draft/multiline extension.

### DIFF
--- a/modules/core/m_message.c
+++ b/modules/core/m_message.c
@@ -34,6 +34,7 @@
 #include "channel.h"
 #include "match.h"
 #include "hash.h"
+#include "batch.h"
 #include "class.h"
 #include "msg.h"
 #include "packet.h"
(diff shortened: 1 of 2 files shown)


skizzerz:
Note: I'm also fine sending up the full multiline PR (which could include this) and closing out this one, if you want to see how it's actually used.

aaronmdjones:
That would be ideal.  I don't know whether this is missing a check to prohibit clients from opening this batch type for example.

PR #523, opened 2026-05-13
Remove some legacy cruft

skizzerz:
- The hurt extension doesn't work well and is now supplanted by m_quarantine
- ip_cloaking_old has been broken (causes segfaults) for years now, remove it
- ip_cloaking_3.0 has hardcoded keys that require .c file edits to be secure. Anyone using it will probably maintain their fork of this file anyway, but it's not good for new people to use. Remove that too
- Remove LOCOPS and umode +l entirely; they serve no useful purpose on unified networks. This also removes the no_locops extension since it is now useless


ircd/s_conf.c
```diff
@@ -780,8 +780,7 @@ set_default_conf(void)
 	ConfigFileEntry.tls_ciphers_oper_only = false;
 	ConfigFileEntry.oper_secure_only = false;
 
-	ConfigFileEntry.oper_umodes = UMODE_LOCOPS | UMODE_SERVNOTICE |
-		UMODE_OPERWALL | UMODE_WALLOP;
+	ConfigFileEntry.oper_umodes = UMODE_SERVNOTICE | UMODE_OPERWALL | UMODE_WALLOP;
```
aaronmdjones: I noticed that there's a `DEFAULT_OPER_UMODES` macro which this PR adjusts to also match this. I'd just use that here.

aaronmdjones (approved):

PR #524, opened 2026-05-13
Add implementation of draft react spec

skizzerz:
For +draft/react, we limit the reaction to a single emoji (potentially including modifiers) for easier moderation (since this doesn't go through spamfilter). Additionally, the +draft/unreact tag was implemented as a means for users to easily remove their reactions. This dependence on emoji brings in a new dependency on libicu.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -324,6 +324,30 @@ AC_CHECK_HEADER(netinet/sctp.h, [
 
 fi
 
+
+AC_ARG_ENABLE(icu,
+AC_HELP_STRING([--disable-icu],[Disable ICU support]),
+[icu=$enableval],[icu=yes])
+
+AS_IF([test "$icu" = yes], [
+	save_CPPFLAGS="$CPPFLAGS"
+	save_LIBS="$LIBS"
+	PKG_CHECK_MODULES([ICU], [icu-uc >= 70], [], [icu=no])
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -324,6 +324,30 @@ AC_CHECK_HEADER(netinet/sctp.h, [
...
+	AC_CHECK_HEADERS([unicode/ustring.h unicode/uchar.h],
+	[
+	    AC_DEFINE(HAVE_LIBICU, 1, [Define to 1 if libicu is available.])
+	], [icu=no])
+])
+
+AS_IF([test "$icu" = no], [
+	CPPFLAGS="$save_CPPFLAGS"
+	LIBS="$save_LIBS"
+])
```
aaronmdjones, line 341: This needs to be at the end of the inside of the AS_IF above or else it will set them to the values used in the start of the hyperscan test (if hyperscan was enabled) or the OpenSSL test (if not).

configure.ac
aaronmdjones, line 296: For that reason I would move these to the end of the inside of the AS_IF immediately above too.

configure.ac
skizzerz: Solved a different way by pulling save_* out of the AS_IF. Seemed cleaner than nested tests.

PR #525, opened 2026-05-13
Add draft/multiline support

skizzerz:
This includes two components:

- A new core s2s batch type solanum.chat/echo controlled by the `ECHOB` server capability. Used for echo-message support for batched messages. Note that unlike normal commands, the contents of an echo batch are sent to the client as-is (e.g. no parsing/substitution of UIDs/SIDs). Limited validation is still performed to ensure the prefix/origin is not coming from a fake direction. The echo batch parameters contain the origin instead which will be sent to the target client rather than any origin inside of the batch messages.
- The draft/multiline extension. This is currently explicitly unstable due to the s2s batch changing from draft/multiline to unprefixed multiline once the spec is ratified. The `MULTILN` server capability will switch meaning as well come ratification time. This is an implementation detail and could be adjusted if desired (to e.g. just use unprefixed multiline for s2s batches even right now which would allow better mixed-version support in the future).

### DIFF
--- a/extensions/multiline.c
+++ b/extensions/multiline.c
@@ -1,0 +1,812 @@
+/*
+ * Solanum: a slightly advanced ircd
+ * multiline.c: Implement the draft/multiline batch type
+ *
+ * Copyright (c) 2026 Ryan Schmidt <skizzerz@skizzerz.net>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "stdinc.h"
+#include "batch.h"
+#include "client.h"
+#include "hash.h"
+#include "modules.h"
+#include "msgbuf.h"
+#include "newconf.h"
+#include "numeric.h"
+#include "send.h"
+#include "s_assert.h"
+#include "s_conf.h"
+#include "s_newconf.h"
+#include "s_serv.h"
+#include "tgchange.h"
+
+/* the MULTILN server capability indicates support for the s2s multiline batch type
+ * since this is a proof-of-concept extension the s2s interface is UNSTABLE and may change incompatibly.
+ * Networks need to ensure all servers have the same module version to avoid potential crashes.
+ */
+static uint64_t CAP_MULTILINE;
+static uint64_t CLICAP_MULTILINE;
+static int h_multiline_message_user;
+static int h_multiline_message_channel;
+
+static const char multiline_desc[] =
+	"Allows clients to send batches containing multiple lines at once (draft/multiline batch type).";
+
+static char capdata_buf[BUFSIZE];
+static unsigned int old_max_lines = 0;
+static unsigned int old_max_bytes = 0;
+static unsigned int max_lines = 0;
+static unsigned int max_bytes = 0;
+
+static int multiline_modinit(void);
+static void multiline_moddeinit(void);
+static void process_multiline(struct Client *client_p, struct Client *source_p, struct Batch *batch, void *);
+static void multiline_conf_info(void *);
+static void multiline_conf_store(void *);
+static void multiline_conf_update(void *);
+static void multiline_tag_allow(void *);
+static void multiline_rehash(void *);
+static const char *multiline_config(struct Client *);
+
+mapi_hfn_list_av1 multiline_hfn_list[] = {
+	{ "conf_read_start", multiline_conf_store },
+	{ "conf_read_end", multiline_conf_update },
+	{ "doing_info_conf", multiline_conf_info },
+	{ "message_tag", multiline_tag_allow },
+	{ "rehash", multiline_rehash },
+	{ NULL, NULL },
+};
+
+mapi_hlist_av1 multiline_hlist[] = {
+	{ "multiline_message_user", &h_multiline_message_user },
+	{ "multiline_message_channel", &h_multiline_message_channel },
+	{ NULL, NULL },
+};
+
+struct ClientCapability capdata_multiline = { .data = multiline_config };
+
+mapi_cap_list_av2 multiline_cap_list[] = {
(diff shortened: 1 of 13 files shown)


progval (commented):
Bugs founds by irctest:

* The batch is sent to clients without the `draft/multiline` cap
* Tags inside the batch should be stripped (eg. `time`)
* A client sending `@batch` with an incorrect id should get `FAIL BATCH INVALID_REFTAG`, but gets nothing. eg.
  ```
  1778706225.616 1 -> S: BATCH +123 draft/multiline #test
  1778706225.616 1 -> S: @batch=231 PRIVMSG #test :hi
  1778706225.616 1 -> S: PING synchronize579081.364567705
  1778706226.618 S -> 1: @time=2026-05-13T21:03:46.617Z :My.Little.Server PONG My.Little.Server :synchronize579081.364567705
  ```
* `draft/multiline-concat` messages with blank content are not rejected (the spec does not explicitly require this, but implicitly recommends it)

skizzerz:
> The batch is sent to clients without the `draft/multiline` cap

~~Can you provide a trace/example of this?~~ Spotted the issue, will fix.

> Tags inside the batch should be stripped (eg. `time`)

This is not required by the specification and is too difficult to implement in solanum. This will be left as-is.

> A client sending `@batch` with an incorrect id should get `FAIL BATCH INVALID_REFTAG`, but gets nothing. eg.

This is not required by the specification. In fact, it violates "Once a client has opened a batch, it MUST NOT send any messages that are not part of the batch". The server is allowed to respond (or not respond) however it wants when a client violates the specification.

The INVALID_REFTAG fail code is for the BATCH command, as it indicates that "all past and future messages in this batch will be ignored". A client that subsequently sends commands with the rejected batch tag will have their commands silently rejected (ignored), which is exactly what solanum does.

> `draft/multiline-concat` messages with blank content are not rejected (the spec does not explicitly require this, but implicitly recommends it)

Can you provide a trace of this? It should've been rejected.

progval:
> Can you provide a trace of this? It should've been rejected.

You're right, it's rejected. I didn't see it because it's only rejected when Solanum gets the `BATCH -`, while Ergo rejects it immediately.

progval:
I just noticed that every command in the multiline batch has a different value for the `@time` tag, eg.

```
1781377075.764 S -> 2: @msgid=1178137707467200684442XAAAAABI3Rlc3Q=;time=2026-06-13T18:57:54.731Z :alice!~username@127.0.0.1 BATCH +6i1Dl1g9jlkvJDE draft/multiline #test
1781377075.764 S -> 2: @batch=6i1Dl1g9jlkvJDE;time=2026-06-13T18:57:54.737Z :alice!~username@127.0.0.1 PRIVMSG #test :hello
1781377075.764 S -> 2: @batch=6i1Dl1g9jlkvJDE;time=2026-06-13T18:57:54.740Z :alice!~username@127.0.0.1 PRIVMSG #test :#how is 
1781377075.764 S -> 2: @batch=6i1Dl1g9jlkvJDE;draft/multiline-concat;time=2026-06-13T18:57:54.741Z :alice!~username@127.0.0.1 PRIVMSG #test :everyone?
```

The spec does not specify which one clients should display next to the message; so we may end up with clients displaying different timestamp for the same message. Can this be solved in Solanum, or should we suggest a spec update?

~~For what it's worth, Unreal has the same time on the inner messages as on the `BATCH +` but a different value for `BATCH -`~~ (EDIT: I was wrong, there is no `@time` on Unreal's `BATCH -`)

skizzerz:
Spec update sounds best here. There is no way to do this in a generic fashion in solanum that works for all batch types (as other server-to-client batch types *do* want inner tags), which means I'd either need to select specific tags to suppress for multiline batches (and other tags won't be suppressed), or each tag module needs to specifically indicate how it works with multiline.

This is a lot of work, and it's fairly ugly, for something that can be solved with one sentence in a spec and that clients will probably end up doing anyway even if the spec isn't updated.

aaronmdjones (approved):
This seems OK, or, as confident as I can be on a 2k LoC change.

extensions/multiline.c
```diff
@@ -0,0 +1,865 @@
...
+	{
+		struct BatchMessage *data = ptr->data;
+		if (rb_strcasecmp(command, data->msg.cmd) != 0)
+		{
+			sendto_one(source_p, ":%s FAIL BATCH MULTILINE_INVALID :multiline batch must only have either PRIVMSG or NOTICE commands",
+				me.name);
+			return;
+		}
+
+		if (irccmp(target, data->msg.para[1]) != 0)
```
progval, line 691: Assertion failure when getting this from a client:

```
1789569691.669 1 -> S: BATCH +123 draft/multiline #test
1789569691.669 1 -> S: @batch=123 PRIVMSG #test :hello
1789569691.669 1 -> S: @batch=123 PRIVMSG
1789569691.669 1 -> S: BATCH -123
```

```
1789569692.724 valgrind solanum: ../ircd/match.c:593: irccmp: Assertion `ss_assert(s2 != ((void *)0))' failed.
1789569692.727 valgrind ==3597388== 
1789569692.727 valgrind ==3597388== Process terminating with default action of signal 6 (SIGABRT): dumping core
1789569692.727 valgrind ==3597388==    at 0x49DD69C: __pthread_kill_implementation (pthread_kill.c:44)
1789569692.727 valgrind ==3597388==    by 0x4988C41: raise (raise.c:26)
1789569692.727 valgrind ==3597388==    by 0x49714AB: abort (abort.c:77)
1789569692.727 valgrind ==3597388==    by 0x497141F: __assert_fail_base.cold (assert.c:118)
1789569692.727 valgrind ==3597388==    by 0x4894C24: irccmp (match.c:593)
1789569692.727 valgrind ==3597388==    by 0x59ACDAA: process_multiline (multiline.c:691)
1789569692.727 valgrind ==3597388==    by 0x4919C08: finish_batch (m_batch.c:211)
1789569692.727 valgrind ==3597388==    by 0x491A5DD: m_batch (m_batch.c:413)
1789569692.727 valgrind ==3597388==    by 0x48A14DA: handle_command (parse.c:318)
1789569692.727 valgrind ==3597388==    by 0x48A0F33: parse (parse.c:225)
1789569692.727 valgrind ==3597388==    by 0x48A0595: client_dopacket (packet.c:377)
1789569692.727 valgrind ==3597388==    by 0x489FA64: parse_client_queued (packet.c:150)
```

extensions/multiline.c
```diff
@@ -0,0 +1,865 @@
...
+		|| IsServerCapable(target_p->from, CAP_ECHOB | CAP_MULTILINE))
+	{
+		char echo_batch_id[16];
+		generate_batch_id(echo_batch_id, sizeof(echo_batch_id));
+		struct MsgTag echo_batch = { "batch", echo_batch_id, CLICAP_BATCH };
+
+		/* echo the incoming batch tag rather than our synthesized one */
+		inner_tags[0].value = batch->tag;
+
+		/* send an echo batch here since the target's server won't send one */
```
progval, line 450: This seems inconsistent. We're in this branch because (among other options) `IsServerCapable(target_p->from, CAP_ECHOB | CAP_MULTILINE)`, which implies the target's server will send an echo

PR #527, opened 2026-05-19
Fix use after free in LIST with labeled-response

skizzerz:
In the event that LIST output is small enough to complete without invoking SAFELIST code (i.e. the buffer doesn't exceed half of the client's SendQ), the stashed outgoing_response_info in struct ListClient is the same pointer as the outgoing_response_info for the LIST command. When we finish sending the LIST, we free the stashed struct ListClient's outgoing_response_info and then restore the outgoing_response_info from the LIST command, and attempt to do things with it. However, that memory has already been freed. Detect the case when the pointers are the same value and properly wipe them both so we no longer attempt to read from it.

### DIFF
--- a/modules/m_list.c
+++ b/modules/m_list.c
@@ -497,10 +497,12 @@ static void safelist_client_release(struct Client *client_p, struct ResponseInfo
 	if (outgoing_response_info != NULL)
 		sendto_one(client_p, ":%s BATCH -%s", me.name, outgoing_response_info->batch);
 
-	free_response_batch(outgoing_response_info);
-	/* outgoing_response_info is now pointing to freed memory, so if orig is the same pointer, wipe it */
+	/* outgoing_response_info is about to be freed, so if orig is the same pointer, wipe it */
 	if (outgoing_response_info == *orig_response_info)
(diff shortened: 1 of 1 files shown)


modules/m_list.c
```diff
@@ -498,6 +498,9 @@ static void safelist_client_release(struct Client *client_p)
 		sendto_one(client_p, ":%s BATCH -%s", me.name, outgoing_response_info->batch);
 
 	free_response_batch(outgoing_response_info);
+	/* outgoing_response_info is now pointing to freed memory, so if orig is the same pointer, wipe it */
+	if (outgoing_response_info == *orig_response_info)
+		*orig_response_info = NULL;
```
aaronmdjones: I would prefer that the nulling is done before the freeing. Using a pointer (even in an equality test) after it has been freed is liable to be undefined behaviour according to pointer provenance rules in later C standards.

```C
    // We are about to free this, so be sure to null out the original pointer if it is equal
    if (outgoing_response_info == *orig_response_info)
        *orig_response_info = NULL;

    free_response_batch(outgoing_response_info);
```

aaronmdjones (approved):

PR #529, opened 2026-05-21
Don't auto-expire SAFELIST batches

skizzerz:
Avoids use-after-free if the client takes more than 10 seconds to fully receive the LIST output.


aaronmdjones (approved):

aaronmdjones:
Well that one is nasty. 😬

PR #531, opened 2026-05-23
Update labeled-response code

skizzerz:
The previous API was too error-prone, as evidenced by the large amount of crash bugs discovered with it. New functions were added to response.h to more easily manage labeled-response state, and some state tracking was moved into core (such as CLICAP_RECEIVE_LABEL, which must only ever be on one client at a time). Because this tracking is now in core, cap_labeled_response needs to be a core module as having that bit unloaded will cause issues. m_batch was also moved into being a core module as more and more core things have depended on batch support existing (and more will depend on it in the future).

m_list no longer re-uses remote response batches for SAFELIST, instead making use of new APIs to suspend a labeled-response batch and more easily resume it.

Batch resumption was made more focused as well, to ensure that we don't accidentally lose batches due to setting outgoing_response_info to NULL. The free_response_batch takes a resumption pointer to avoid cases where we could introduce use-after-free by combining
free+resume_response_batch.

struct ResponseInfo now takes strduped pointers instead of directly holding the buffers, and some fields were merged into a flags field or omitted entirely due to new suspension/resumption support. This will have a minor positive impact on memory usage as we no longer need to allocate buffers for maximum-sized strings for every labeled-response.

### DIFF
--- a/modules/m_list.c
+++ b/modules/m_list.c
@@ -69,13 +69,10 @@ static void list_one_channel(struct Client *source_p, struct Channel *chptr, int
 static void safelist_one_channel(struct Client *source_p, struct Channel *chptr, struct ListClient *params);
 static void safelist_check_cliexit(void *);
 static void safelist_client_instantiate(struct Client *, struct ListClient *);
-static void safelist_client_release(struct Client *, struct ResponseInfo **);
+static void safelist_client_release(struct Client *, struct ResponseInfo *);
 static void safelist_iterate_client(struct Client *source_p);
 static void safelist_iterate_clients(void *unused);
 static void safelist_channel_named(struct Client *source_p, const char *name, int operspy);
-
-static uint64_t restore_global_context(struct Client *);
-static void reset_global_context(struct ResponseInfo *, uint64_t);
 
 struct Message list_msgtab = {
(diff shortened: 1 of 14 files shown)


skizzerz:
PR marked as draft because I want to test this out for a bit on my testnet to ensure there are no remaining crash bugs or memory leaks. Some changes may still occur, but feel free to leave review comments now if you wish.

modules/m_list.c
```diff
@@ -112,30 +109,19 @@ static int _modinit(void)
...
-		}
-		else
-			restore_global_context(source_p);
-
+		struct ResponseInfo *orig_info = NULL;
+		if (MyClient(source_p))
+			resume_response_batch(source_p->localClient->safelist_data->response_info);
 		sendto_one_notice(source_p, ":/LIST aborted");
-		safelist_client_release(source_p, &orig_outgoing_response_info);
+		safelist_client_release(source_p, orig_info);
```
aaronmdjones, line 122: Why are you creating a NULL variable here instead of just passing a literal NULL?

skizzerz: Fixed, was missing setting the value above on line 120.

aaronmdjones (approved):

PR #532, opened 2026-05-23
Prevent users from running CAPAB pre-registration

skizzerz:
We now disconnect the client if they attempt to run CAPAB before running PASS (introducing themselves as a TS6 server with an SID) or if they've already sent USER/NICK info. USER/NICK already has the other half of the requisite checks by disconnecting the user if they have an SID. As such, the logic flow of requiring an SID before CAPAB means users can never have server caps set on them.

This prevents various issues such as a client sending CAPAB STAG receiving all message tags, including those designed to be s2s-only.

### DIFF
--- a/modules/m_capab.c
+++ b/modules/m_capab.c
@@ -64,8 +64,17 @@ mr_capab(struct MsgBuf *msgbuf_p, struct Client *client_p, struct Client *source
 	if(client_p->localClient == NULL)
 		return;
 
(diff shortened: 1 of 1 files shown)


aaronmdjones (approved):

aaronmdjones:
Merged in commit dfe6757a577109d2180465da6270b5a6bc08f8d7

GitHub wouldn't let me.

PR #536, opened 2026-05-24
Meson: Decouple creation and generation data from version

aaronmdjones:
This fixes the generation counter being hardcoded at 0 and the IRCd birth date being reported as `__DATE__ at __TIME__`.


skizzerz (approved):
Confirmed this works on my testnet, should be good to merge

spb (approved):

PR #543, opened 2026-07-15
Pass block name to conf_start callbacks

skizzerz:
Configuration blocks of the form `block "name" { ... }` previously were unsupported in extensions calling add_top_conf because the name portion was not exposed to anything within the extension. Pass the name along so callbacks can use that data instead of relying on global state in newconf.c with static storage duration.

### DIFF
--- a/ircd/newconf.c
+++ b/ircd/newconf.c
@@ -92,7 +92,7 @@ conf_strtype(int type)
 }
 
 int
-add_top_conf(const char *name, int (*sfunc) (struct TopConf *),
+add_top_conf(const char *name, int (*sfunc) (struct TopConf *, const char *),
 		int (*efunc) (struct TopConf *), struct ConfEntry *items)
 {
 	struct TopConf *tc;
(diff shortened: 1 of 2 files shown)


ircd/newconf.c
```diff
@@ -1768,12 +1788,12 @@ conf_set_service_name(void *data)
...
 static int
-conf_begin_alias(struct TopConf *tc)
+conf_begin_alias(struct TopConf *tc, const char *name)
 {
 	yy_alias = rb_malloc(sizeof(struct alias_entry));
 
-	if (conf_cur_block_name != NULL)
-		yy_alias->name = rb_strdup(conf_cur_block_name);
+	if (name != NULL)
+		yy_alias->name = rb_strdup(name);
```
aaronmdjones, line 1796: This change seems superfluous? The unit-scope variable is still required for elsewhere in this file (e.g. in `conf_set_privset_privs()`) so we may as well keep using it.

skizzerz: Yes it's a superfluous change, but it also gives an example someone could look up to figure out how the name parameter is actually used, if they want to replicate it in a different topconf block (whether in core or in an extension).

aaronmdjones (approved):

PR #545, opened 2026-07-15
Expand monitor

skizzerz:
This set of changes allows other modules to better interact with a user's MONITOR list via a new hook and is_monitoring function. Some new sendto functions were added to send to an rb_dlink_list of local clients, and sendto_monitor was refactored to use that same underlying machinery. There are no current callers of the new functions, but it will be needed for METADATA support in the future.


aaronmdjones (approved):

jillest:
The context is unclear to me. What are these "other modules" going to do?

skizzerz:
> The context is unclear to me. What are these "other modules" going to do?

The draft/metadata spec requires that we send out metadata batches for online users whenever someone adds a user to their MONITOR list. Additionally, `METADATA *ALL SYNC` (from draft/metadata-3) will synchronize metadata for all visible targets, including online users on the sender's MONITOR list.

PR #547, opened 2026-07-15
Sort channel members in alphabetical order

skizzerz:
Previously, the ordering of members in a channel was based on the reverse ordering in which they joined (i.e. most recent first). While this worked just fine, the lack of a consistent ordering means that things like NAMES and WHO output are arbitrary and (more importantly) that it was impossible to implement asynchronous METADATA listings.

A user updating their nickname or a server forcing a nickchange will re-sort the user's position in the channel list to ensure the ordering remains alphabetical.

As part of this, a handful of cases that duplicated the code to clear a user's accept list were replaced by the helper function that did the same.


skizzerz:
Going to do a different approach on this one

PR #549, opened 2026-08-11
Add optional fallback for +reply tag

skizzerz:
If enabled, we accept incoming +draft/reply tags in addition to +reply (if both are specified, then we read +reply and ignore +draft/reply). Additionally, if either +reply or +draft/reply were given to us, we will provide both +reply and +draft/reply in the outgoing message, with the same value. This fallback is disabled by default but can be enabled via a new ircd.conf option.

A second configuration option (also disabled by default) advertises draft/reply as allowed in CLIENTTAGDENY. The tag's presence in CLIENTTAGDENY bears no impact on whether it is accepted by the software, and networks may wish to keep it off the list in case it is already very long or if they want to encourage clients to upgrade to the non-draft spec.

### DIFF
--- a/extensions/tag_reply.c
+++ b/extensions/tag_reply.c
@@ -32,6 +32,7 @@
 #include "numeric.h"
 #include "chmode.h"
 #include "hash.h"
+#include "newconf.h"
 #include "parse.h"
 #include "inline/stringops.h"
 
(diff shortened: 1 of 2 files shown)


extensions/tag_reply.c
```diff
@@ -140,4 +176,53 @@ tag_reply_allow(void *data_)
...
+	else
+		remove_client_tag("draft/reply");
+}
+
+static void
+tag_reply_reset_conf(void *unused)
+{
+	enable_fallback = 0;
+	advertise_fallback = 0;
+}
```
aaronmdjones, line 193: This is missing a call to `remove_client_tag("draft/reply");`

The reason it is necessary is that `add_client_tag` does not guard against duplicate entries. Rehashing multiple times with both options enabled will add it multiple times.

I suggest moving it from the function above to here.

PR #556, opened 2026-09-11
Check for IsOper instead of IsOperGeneral for command handlers

aaronmdjones:
Using `IsOperGeneral` to decide whether to use the normal or oper command handler for a client means that opered-up network bots and such will not be able to execute oper commands unless they have the `oper:general` privilege granted to them.  This privilege allows them to do far more than execute oper commands.


dwfreed:
I wonder why `oper:general` was picked in the first place...

skizzerz:
Things this now gives all opers access to that previously implicitly required `oper:general`:

- `SQUIT` for a locally-connected server
- `CLOSE` (terminates all local connections that have not yet finished registration)
- `CONNECT` for direct (non-remote) links
- `ETRACE`
- `MASKTRACE`
- `SCAN`
- `SET` (with exception of `SET IDENTTIMEOUT` which requires `oper:admin`)
- `TESTMASK`
- `UNREJECT`
- `OKICK`

Other behavioural changes that previously implicitly required `oper:general` and are now available to all opers:

- Removal of rate limiting for `ADMIN`
- Receiving oper help files when executing `HELP`
- Removal of rate limiting for `INFO` (listing conf options in `INFO` still requires `oper:general` however)
- Viewing full `LINKS` when server hiding is enabled
- Removal of rate limiting for `LIST`
- Removal of rate limiting for `LUSERS`
- Viewing full `MAP` when server hiding is enabled and viewing missing (recently disconnected) servers in `MAP`
- Removal of rate limiting for `MOTD` on remote servers (there was already no rate limiting for parameterless `MOTD`)
- Removal of rate limiting for `VERSION`
- Removal of rate limiting for `MKPASSWD`

PR #557, opened 2026-09-11
Make CONNECT and SQUIT always require oper:routing

supertassu:
All network routing changes should require oper:routing. Previously
these have only implicitely required oper:general to select the command
handler type.


aaronmdjones:
I'm already working on this right now.

PR #558, opened 2026-09-12
Fix tag_channel_context

skizzerz:
Only process incoming tags if they're +channel-context


aaronmdjones (dismissed):

aaronmdjones (approved):

skizzerz:
This has been tested and should be good to merge now.

